PR #45, opened 2024-03-24
Fix CUDA build on Rust 1.77

zeux:
- CUDA Compiler trait impls didn't define Output type so --features cuda wouldn't compile
- nn module wasn't exporting submodules as public so imports like this in Mistral demo didn't resolve:

```
use luminal::{
    nn::{embedding::Embedding, norm::RMSNorm},
```

After these fixes, `cargo run --release --features cuda` works from examples/mistral on Rust 1.77.


jafioti:
@zeux  Looks good! Only nitpick I have is changing all the nn modules to pub. Can you instead just change this:
```
use luminal::{
    nn::{embedding::Embedding, norm::RMSNorm},
```
to
```
use luminal::prelude::*;
```
because all the nn modules are re-exported by nn, which should be covered by prelude.

zeux:
Yeah definitely, I can fix the import side - wasn't sure which direction to go.

jafioti:
Looks great, thanks!

Issue #64, opened 2024-06-07
Allow noop `expand`s


swfsql:
Hi, is it possible to allow noop expands?

Example:
```rust
#[test]
fn noop_expand() {
    let cx = Graph::new();
    let a = cx.tensor::<R1<1>>();
    let mut b = a.clone();
    b = a.expand::<R1<1>, Axis<0>>();
}
```
Note: the example compiles successfully for `R0`, as the noop reduce/broadcast is allowed for this case.

This results in the following compilation error:
```
error[E0277]: the trait bound `(luminal::shape::Const<1>,): luminal::shape::BroadcastShapeTo<(luminal::shape::Const<1>,), _>` is not satisfied
   --> src/main.rs:7:15
    |
102 |         b = a.expand();
    |               ^^^^^^ the trait `ReduceShapeTo<(luminal::shape::Const<1>,), _>` is not implemented for `(luminal::shape::Const<1>,)`, which is required by `(luminal::shape::Const<1>,): luminal::shape::BroadcastShapeTo<_, _>`
    |
    = help: the following other types implement trait `ReduceShapeTo<S, Ax>`:
              <() as ReduceShapeTo<(), Axis<0>>>
              <(A, B) as ReduceShapeTo<(), Axes2<0, 1>>>
              <(A, B) as ReduceShapeTo<(A,), Axis<1>>>
              <(A, B) as ReduceShapeTo<(B,), Axis<0>>>
              <(A, B, C) as ReduceShapeTo<(), Axes3<0, 1, 2>>>
              <(A, B, C) as ReduceShapeTo<(A, B), Axis<2>>>
              <(A, B, C) as ReduceShapeTo<(A, C), Axis<1>>>
              <(A, B, C) as ReduceShapeTo<(A,), Axes2<1, 2>>>
            and 113 others
    = note: required for `(luminal::shape::Const<1>,)` to implement `luminal::shape::BroadcastShapeTo<(luminal::shape::Const<1>,), _>`
note: required by a bound in `movement::<impl luminal::graph_tensor::GraphTensor<S>>::expand`
   --> luminal/src/hl_ops/movement.rs:17:12
    |
15  |     pub fn expand<Dst: Shape, Ax: Axes>(mut self) -> GraphTensor<Dst>
    |            ------ required by a bound in this associated function
16  |     where
17  |         S: BroadcastShapeTo<Dst, Ax>,
[...]

PR #67, opened 2024-06-17
allow noop tensor `expand`

swfsql:
Example related to #64


swfsql:
I'm re-opening because there is a simpler test, and also because if the PR is closed no new commits appear on it (which makes sense).

swfsql:
I added a change that checks if src == dst during expansion, and stops the actual shape modifications in case it is.
But although the small test now passes, I still have some shape error in another test which is not a minimal example, so I'll keep trying to reduce or notice what is the error

jafioti:
@swfsql I've just pushed a really large change which moves away from const generic shapes and to only using ShapeTracker runtime shape tracking. That means this PR is probably not nessecary anymore since we don't have a .expand() anymore, it's now .expand(axis, size).

I'm still interested in const generic shape tracking, but currently rust doesn't have enough support for const generic expressions, which made the code very complex (as you're seeing) and some models unrepresentable. Once better support on the rust langauge level is stablized, we can bring it back in.

Issue #68, opened 2024-06-20
Autograd add grad error


swfsql:
Hi, I've found a runtime error and made this minimized example:

```rust
#[test]
fn test_add_grad_decreasing_idx() {
    let mut cx = Graph::new();
    let a: GraphTensor<R1<2>> = cx.tensor();
    let a: GraphTensor<R3<1, 1, 2>> = a.expand::<_, LAxes2<0, 1>>();
    let a: GraphTensor<R3<2, 1, 1>> = a.permute::<_, LAxes3<2, 1, 0>>();
    // a.shape.fake = [false, true, true]
    // a.shape.indexes = [0, 2, 1] // note that the idx isn't necessarily increasing (0,1,2)
    let b: GraphTensor<R3<2, 1, 1>> = cx.tensor();
    let weights = vec![a.id, b.id];

    let m: GraphTensor<R3<2, 1, 1>> = a * b;
    let loss: GraphTensor<R0> = m.sum_reduce();
    let _grads = cx.compile(Autograd::new(weights, loss), ());
}
```

The error:
```
thread 'autograd::tests::test_add_grad_decreasing_idx' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/tinyvec-1.6.0/src/arrayvec.rs:681:26:
index out of bounds: the len is 0 but the index is 0
stack backtrace:
   ..
   3: tinyvec::arrayvec::ArrayVec<A>::remove
             at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/tinyvec-1.6.0/src/arrayvec.rs:681:26
   4: luminal::shape::tracker::ShapeTracker::remove_dim
             at /workspaces/coursera-deep-learning-specialization/luminal_original/src/shape/tracker.rs:58:21
   5: luminal_training::autograd::add_grad
             at ./src/autograd.rs:208:13
   6: <luminal_training::autograd::Autograd as luminal::compiler_utils::Compiler>::compile
             at ./src/autograd.rs:113:21
   ..
```

[...]

PR #69, opened 2024-06-20
Autograd add grad error (failing test)

swfsql:
Related to #68

### DIFF
--- a/crates/luminal_training/src/autograd.rs
+++ b/crates/luminal_training/src/autograd.rs
@@ -552,4 +552,20 @@ mod tests {
                 .as_vec(),
         );
     }
+
+    #[test]
+    fn test_add_grad_decreasing_idx() {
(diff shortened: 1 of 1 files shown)


swfsql:
Note: reviewing the fix code, I think the it's still not correct because ultimately every different index may have it's own index diff that precedent removals may have interfered with it. Currently, only the indexes above the "global minimal" get a index diff, but not all the would need to.

swfsql:
Another note to add is that although the status quo (`main`) code doesn't fail on the second test (`test_add_grad_decreasing_idx_r2`), I think it removes the wrong dimension. This would happen because the vector still has enough items, but the wrong item would get removed.

jordanmb:
I tried applying this patch and it does not fix the issues for my workflow. I was really hoping this library would be usable because I love the approach.

Issue #70, opened 2024-06-22
Cuda `GatherCompiler` fails on low dimensionality


swfsql:
Hi, I've encountered a failure on Cuda's `GatherCompiler` compilation pass.

I've extracted the one_hot functionality from embeddings and applied it to a transformation of some input tensor. It succeeds if the input is a `R1`, but fails if it's a `R0`.

I've added a PR containing an example of a failing test.

jafioti:
Solved with #71

PR #71, opened 2024-06-22
Cuda `GatherCompiler` fails on low dimensionality (failing test)

swfsql:
Related to #70.

This showcases a failure for the Cuda `GatherCompiler` compilation pass.

To fetch the PR and run the tests:
```bash
git fetch origin pull/71/head:cuda_gather_failure
git switch cuda_gather_failure
cd crates/luminal_cuda
cargo test test_gather_compiler
```

The `test_gather_compiler_r1` passes:
https://github.com/jafioti/luminal/pull/71/files#diff-ead6d80249befe24da4ea23c236763f82cc1eccfccbb92035217e620e8bf5184R434-R459

While `test_gather_compiler_r0` fails:
https://github.com/jafioti/luminal/pull/71/files#diff-ead6d80249befe24da4ea23c236763f82cc1eccfccbb92035217e620e8bf5184R409-R432

The failure error is:
```
---- binary::tests::test_gather_compiler_r0 stdout ----
thread 'binary::tests::test_gather_compiler_r0' panicked at src/binary.rs:379:46:
index out of bounds: the len is 2 but the index is 2
```

The relevant line of code is:
https://github.com/jafioti/luminal/blob/f61d53f859293d4c5e4d57d264dfc423a98007e0/crates/luminal_cuda/src/binary.rs#L379

I'm not sure how cuda relates to those operations, but the first thought would be to maybe take the last dimension value instead of the third (`emb_shape.shape()[2]`) one?

### DIFF
--- a/crates/luminal_cuda/src/binary.rs
+++ b/crates/luminal_cuda/src/binary.rs
@@ -376,7 +376,7 @@ impl<T: CudaFloat> Compiler for GatherCompiler<T> {
                 .as_data()
                 .unwrap()
                 .2;
-            let embed_dim = emb_shape.shape()[2].to_usize().unwrap();
+            let embed_dim = emb_shape.shape().last().unwrap().to_usize().unwrap();
             let index_shape = graph
                 .edges_connecting(s.get(&indexes), s.get(&ind_copy))
                 .next()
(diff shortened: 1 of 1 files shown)


jafioti:
Yeah you're right, taking the last dimension here was the correct solution. Can you just change that line to
```
let embed_dim = emb_shape.shape().last().unwrap().to_usize().unwrap();
```
That should get the test passing, and then I'll merge

swfsql:
Thanks, I've added this change, and the tests pass.

jafioti:
Great thanks!

PR #76, opened 2024-07-27
Fix clip operation on tensor and add test for it.

janroden:
See title. Assuming that the `clip` operation is supposed to work [like in PyTorch](https://pytorch.org/docs/stable/generated/torch.clip.html).

BTW: might be better to rename it to `clamp`, because it seems more commonly used?


jafioti:
Good catch!

PR #79, opened 2025-04-22
adding example documentation for llama 3b inferencing

matthewjgunton:
Adds in a detailed explanation of the Llama 3 8B inferencing example


jafioti:
thanks!

PR #92, opened 2025-06-09
unsqueeze function added

matthewjgunton:


jafioti:
check clippy errors and i think unsqueeze should just be self.expand(dim, 1)

matthewjgunton:
clippy errors already fixed
expand is a bit cleaner than reshape, I'll put out a new commit

PR #159, opened 2025-09-12
Fix add_grad logic and add tests for non-monotonic indexes mapping

### LINKED ISSUE #68: Autograd add grad error

SinanGncgl:
Fix add_grad logic for fake dimensions and add tests for non-monotonic index mappings. I also ran and tested several models, everything seems to work, but another review would be appreciated.

Closes #68

### DIFF
--- a/crates/luminal_training/src/autograd.rs
+++ b/crates/luminal_training/src/autograd.rs
@@ -199,7 +199,9 @@ fn add_grad(
     grad.shape.indexes = new_indexes;
 
     // Undo expands (sum reduce)
-    for i in fwd.shape.indexes.into_iter().rev() {
+    // Iterate over logical dimension indices in reverse order to correctly identify fake dimensions.
+    // The fake dimension flags are stored in logical dimension order, not physical storage order.
+    for i in (0..fwd.shape.len()).rev() {
         if fwd.shape.fake[i] {
             grad.id = graph
                 .add_op(SumReduce(i))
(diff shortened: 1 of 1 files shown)


jafioti:
wait does this work? fakes should be stored in physical order, not logical order, which is why we go through indexes first. if that works there must be a bug somewhere else

SinanGncgl:
@jafioti, that was my first impression: "could not be that easy," but it seems to work.

SinanGncgl:
I am closing this, seems like this is not gonna get somewhere.

PR #161, opened 2025-09-14
OpenCL backend primitive compiler

vasucp1207:
This only implements the primitive operations using rust ocl crate. More to come in the future. Related to https://github.com/luminal-ai/luminal/issues/146

f16 support is not currently available because the ocl::OclPrm trait does not implement f16.

```bash
cx.compile(OpenCLCompiler::<f16>::default(), &mut b);
```
 Maybe we need a custom wrapper for solving this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ exclude = [
     "examples/yolo_v8",
     "crates/luminal_cuda",
     "crates/luminal_metal",
+    "crates/luminal_opencl",
     "crates/luminal_2",
     "demos/flash_attention",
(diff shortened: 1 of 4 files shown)


jafioti:
incredible work! i'll take a look and have it merged soon. in a week or so we'll have opencl and vulkan merged

jafioti:
We've had a ton of churn in the compiler and switched to an egglog-centric arch, so unfortunately there's too much of a gap to merge this. If you have any interest in picking this up and updating it to fit into the modern arch, I'd love to get it in

PR #168, opened 2025-09-30
Added CUDA CI workflow

jayshah1819:
Added CI for CUDA kernels with Docker support. The workflow runs formatting checks, clippy, and tests for the quantized matrix operations (int8, fp16, fp32). Requires a self-hosted runner with NVIDIA GPU to actually run the tests.


jafioti:
How can this be integrated with our current CI on github actions?

PR #170, opened 2025-10-07
adam optimizer implementation

danielleiszen:
Since m and v are defined by a recursive formula, I decided to store the state for the optimizer, which can be updated after each execution step by calling the step_after_execution method. The state is stored for each gradient in AdamGradientState which is collected in the AdamOptimizer. So unlike the SDG implementation not all of the state is stored in the graph. If you suggest a better way of doing this, I am happy to refactor.

### DIFF
--- a/crates/luminal_training/src/optimizer.rs
+++ b/crates/luminal_training/src/optimizer.rs
@@ -58,5 +58,252 @@ pub fn sgd_on_graph(
     (new_weights, lr)
 }
 
-// /// Implements the [Adam](https://arxiv.org/abs/1412.6980) algorithm.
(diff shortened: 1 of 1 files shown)


jafioti:
This is awesome! Can you get the fmt check to pass and then I'll merge?

danielleiszen:
@jafioti 
I modified the solution to use node ids and recreate the tensors when needed. This may be better in certain cases. I fixed the fmt test also. I also tried to test it against the burn implementation, but I suspect the autograd logic is slightly different, because I got different results. I don't know if being numerically equivalent with other libraries is a requirement. Let me know, if I should modify anything else.

jafioti:
Looks great! Merged

PR #171, opened 2025-10-16
Minor cleanups

Cobord:
- Fix up some of the doc comments (backticks)
- On options map and then unwraps are replaced with and_then, is_some_and, ...
- Be more explicit about casts to be clear when as would have been fallible vs infallible
- Panics and Error docs in format so they show up as their own sections
- Direct comparisons with Some(0)... instead of mapping and unwrapping
- Question about the outputs of .compile() never being used, would expect the Looped<C> to use have Self::Output built from the C::Output(s)
-  Instead of directly accessing graph.graph.node_weight, etc inspecting in fields of the struct have methods to do these common tasks. They are inlined methods on Graph instead.
- Tests assert in addition to printing. Want it to be testing properly even if don't --capture output
- A disjoint union of Graph and a test thereof. There are further things to discuss about GraphTensor with this point.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -8,8 +8,12 @@ use std::{
 use super::compiler_utils::{ToIds, ToIdsMut};
 use colored::Colorize;
 use itertools::Itertools;
-use petgraph::{stable_graph::StableGraph, visit::EdgeRef, Direction};
-use rustc_hash::{FxHashMap, FxHashSet};
+use petgraph::{
+    stable_graph::StableGraph,
+    visit::{EdgeRef, IntoEdgeReferences},
+    Direction,
+};
+use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
 
 pub type StorageGraph = StableGraph<Box<dyn Operator>, Dependency>;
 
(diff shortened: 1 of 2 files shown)


jafioti:
What was cleaned up? Please provide details, there are many lines changed

Cobord:
Not too much to say, but I will do so either tomorrow or Monday after can get back to this. Should have put as draft because didn't get to the description part yesterday before context switching.

Cobord:
Any more documentation of what was done that you suggest?

Cobord:
Let me know if there is any more wanted written here. This one is mostly about making more idiomatic. I find that once that happens it is much easier for me to add features without having the linter full of extraneous complaints.

jafioti:
I don't think so, I just havent had a chance to review it yet because a good amount of lines were changed. Should have time today or this week!

Cobord:
Whatever happened here?

jafioti:
> Whatever happened here?

We got a little ambitious with our rewrite and it took longer than expected and changed a lot of the api surface. Most of this pr is still applicable though so I'll be merging it in sometime in the next week or so.

jafioti:
most of these apis have been removed / changed in the past few months now so i'm going to close this. however the api should be much closer to stable at this point, so if you find opportunities for similar cleanups in the current code, please open a pr

PR #176, opened 2025-10-19
Create gemm.rs

salavathhari:


salavathhari:
i created gemm.rs

jafioti:
cuda and metal code should not live in the core

PR #177, opened 2025-10-23
init: rewrite Rust API

MilkBlock:
as title

### DIFF
--- a/crates/luminal_2_eggplant/.gitignore
+++ b/crates/luminal_2_eggplant/.gitignore
@@ -1,0 +1,2 @@
+/target
+*.dot
--- a/crates/luminal_2_eggplant/Cargo.toml
+++ b/crates/luminal_2_eggplant/Cargo.toml
@@ -1,0 +1,54 @@
+[package]
+name = "luminal_2_eggplant"
+version = "0.1.0"
+edition = "2024"
+rust-version = "1.89"
+
+[features]
+default = []
+cuda = ["dep:cudarc", "dep:luminal_cuda"]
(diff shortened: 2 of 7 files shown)


jafioti:
awesome! is there a reason we need an "Unknown" variant for the base types?

MilkBlock:
> is there a reason we need an "Unknown" variant for the base types?

This will be removed in next several versions. Current implementation depends on Default trait and this can be avoided by additional coding but not my first priority.

I'm currently thinking about how to describe tensor core pattern by generic type like
```rust
add_rule(
||{
    LoopOutPat<LoopInPat<XXX>,XXX>::new(xxx)
},
|ctx,pat|{
   LoopOutPat<LoopInPat<XXX>,XXX>:::action(ctx,pat)
}
```
which is more engineering readable.

jafioti:
@MilkBlock is the goal behind eggplant to primarily be a egglog rust api, or do you want it to do other things? I'd like to use it as a part of luminal if I can keep  the main dependency on egglog and eggplant is just a rust frontend.

MilkBlock:
Yeah I can ensure that eggplant will only support egglog backend. And basing on egglog backend eggplant provides different egraph variants as plugins, which are implemented by hook mechanism provided by macro like `#[dsl]`, you can import any of them like changing the `dsl` to `slotted_dsl`. The vanilla egraph (without any plugin or we call pure egglog frontend) is quite stable now, I'm not going to do something different from egglog backend on that.

jafioti:
currently we're experimenting with a simplified thin api directly on top of the egglog crate, and working with the egglog team to directly expand their rust api

PR #180, opened 2025-10-31
Fix egglog compilation error for ShowHN demo

### LINKED ISSUE #179: ShowHN Example Build Failure
```
To run the demo shown on HN on mac, clone this repo and run:

cd demos/matmul
cargo run --release --features metal
```

I tried to build this example, and `egglog` build failed:
```
[ashepelev@thinkpadx1-carbon matmul]$ cargo run --release metal
   Compiling egglog v1.0.0 (https://github.com/MilkBlock/egglog?branch=main#d3c80a49)
error[E0063]: missing field `extra` in initializer of `ClassData`
   --> /home/ashepelev/.cargo/git/checkouts/egglog-c3f6fc3af0f1f45d/d3c80a4/src/serialize.rs:375:13
    |
375 |             egraph_serialize::ClassData {
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `extra`

For more information about this error, try `rustc --explain E0063`.
error: could not compile `egglog` (lib) due to 1 previous error
[ashepelev@thinkpadx1-carbon matmul]$ 
```

For reference, the llama example doesn't generate this error (on my GPU-less machine only says `Either metal or cuda feature must be used for this example!`), so I suspect the problem is with some dependencies in the build

amemov:
Closes #179 
- Removed unused egglog_proof dependency  
- Use egraph-serialize 0.3.0 from crates.io instead of git
- Use egglog 1.0.0 from crates.io instead of git
- Fix `Span` namespace conflict between `ratatui` and `egglog`
- Add `RustSpan` and `Span` imports from `egglog::prelude`

The demo now builds successfully with `--features cuda` & `--features metal`. 

Before merging, it would be a good idea to validate that nothing got broken for either target (I don't have access to GPU), but at least now no compilation errors

### DIFF
--- a/crates/luminal_2/Cargo.toml
+++ b/crates/luminal_2/Cargo.toml
@@ -20,7 +20,7 @@ cudarc = { version = "0.16.6", features = [
 objc2 = { version = "0.6.2", optional = true }
 objc2-metal = { version = "0.3.1", optional = true }
 itertools = "0.14.0"
-egglog = { git = "https://github.com/egraphs-good/egglog", branch="main" }
+egglog = "1.0.0"
 egglog-ast = { git = "https://github.com/egraphs-good/egglog", branch="main" }
 urlencoding = "2.1.3"
 webbrowser = "1.0.4"


jafioti:
Been meaning to update the deps to use egglog 1.0 and egraph serialize to 0.3.0. If you can change those I'll merge

amemov:
weird_bug_matmul_cuda

> ShowHN demo 

For some unknown reason, I had a weird bug above on my machine with CUDA. Not 1000% sure if it is not due to WSL, or something related to Nvidia drivers, or the code - I will try to check that at the end of the week, when I will be installing Linux on the main machine. I have the same problem with `llama` example, so my guess is it's something WSL/drivers related. 

I pushed the commit that changes egglog to 1.0.0

PR #182, opened 2025-11-16
Fix autograd gradient normalization for permuted and broadcast tensors

### LINKED ISSUE #68: Autograd add grad error

ppaleja:
## Summary
Fixes #68 - resolves an index out-of-bounds panic in autograd's add_grad function that occurred when computing gradients for tensors with non-monotonic permute indices and fake (broadcasted) dimensions.

### Remarks
I do admit these fixes seem extremely hacky to me. Would appreciate review and even if they work, it might be evidence for the brittleness of ShapeTracker/autograd logic in its current state. 
I also came across another issue with `permute` logic during my debugging, but will make a separate issue for that.

## Commits
The fix is structured across 4 commits for easier cherrypicking:

### 1. Add tests for identifying bug in issue #68 (8c1b5a7)
Reorganized test module structure for better organization
Added comprehensive test cases that reproduce the issue, including:
test_add_grad_decreasing_idx: The original failing case from issue #68
test_add_grad_high_rank: Tests gradient computation with higher-rank tensors
test_add_grad_preserves_unfaked_dim: Tests preservation of non-broadcast dimensions

Status: All new tests fail at this commit (as expected)

### 2. Refactor gradient normalization logic (777d4b2)
Extracted gradient normalization into normalize_grad_for_input helper function
No behavioral changes, pure refactoring
Added unit tests specifically for the normalization logic
Makes the subsequent fixes clearer and more testable

Status: Tests still fail

### 3. Fix add_grad fake-dim removal (Partial fix) (6e898fd)
Core fix: Changed from using logical indices (fwd.shape.indexes) to physical indices when identifying and removing fake dimensions
Now correctly collects fake underlying indices: (0..fwd.shape.len()).filter(|&idx| fwd.shape.fake[idx])
Sorts indices in descending order before removal to prevent index shifting issues
Why this works: Eliminates the mixing of logical/physical index spaces that caused incorrect axis access

Status: Fixes the OOB panic from the original issue, but test_add_grad_preserves_unfaked_dim still fails because rank alignment isn't handled yet

### 4. Fix broadcasted axes in autograd gradient norm (Complete fix) (4057c7e)
Handles the case where incoming gradients have fewer axes than the forward tensor
Inserts missing axes at positions where the forward tensor has fake dimensions before performing undo-permute logic
Uses grad.expand() to ensure gradient shape aligns with forward tensor shape

Remarks: I think this is the hackiest part of the changes, but it does seem necessary to pass `test_add_grad_preserves_unfaked_dim` (which I verified separately is valid logic in pytorch). Unclear if there is a better way.

Status: All tests pass

### DIFF
--- a/crates/luminal_training/src/autograd.rs
+++ b/crates/luminal_training/src/autograd.rs
@@ -189,6 +189,34 @@ fn normalize_grad_for_input(
     graph: &mut Graph,
 ) -> GraphTensor {
 
+    // If grad has fewer logical axes than fwd (common when forward used expand/broadcast),
+    // insert broadcasted axes into grad at the logical positions where fwd is fake.
+    // Do this before undoing permutes so indexes line up.
+    if grad.shape.len() < fwd.shape.len() {
+        let _need = fwd.shape.len() - grad.shape.len();
+        // insert fake positions from highest -> lowest so indices remain valid
+        let mut fake_positions: Vec<usize> =
+            (0..fwd.shape.len()).filter(|&i| fwd.shape.fake[i]).collect();
+        fake_positions.sort_unstable_by(|a, b| b.cmp(a));
+        for &pos in &fake_positions {
+            if grad.shape.len() >= fwd.shape.len() {
+                break;
+            }
+            // Insert a logical axis with the forward logical size (broadcast view).
+            let size = fwd.shape.dims[fwd.shape.indexes[pos]];
+            grad.shape.expand_dim(pos, size);
+            // Rewrap so the updated shape is used downstream.
+            grad = GraphTensor::from_id(grad.id, grad.shape, graph as *mut Graph);
+        }
+        // Defensive: if we still didn't reach target rank (e.g., malformed fake flags), pad at end.
+        while grad.shape.len() < fwd.shape.len() {
+            let pos = grad.shape.len();
+            let size = fwd.shape.dims[fwd.shape.indexes[pos]];
+            grad.shape.expand_dim(pos, size);
+            grad = GraphTensor::from_id(grad.id, grad.shape, graph as *mut Graph);
+        }
+    }
+
     // Undo permutes
     let mut new_indexes = ArrayVec::new();
     new_indexes.resize(fwd.shape.len(), 0);
--- a/crates/luminal_training/src/autograd.rs
+++ b/crates/luminal_training/src/autograd.rs
@@ -242,313 +242,734 @@ mod tests {
     use luminal::prelude::Module as LModule;
     luminal::test_imports!();
 
+    // Helper utilities used by tests
     fn get_vec(grad: (NodeIndex, ShapeTracker), cx: &mut Graph) -> Vec<f32> {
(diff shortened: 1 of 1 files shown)


jafioti:
great work on this. we've temporarily removed luminal_training, should be back in a month or so

PR #192, opened 2025-12-27
Egglog Expressions

jafioti:
Switched expressions to use egglog instead of egg, and directly use the same MExpr logic as used in full search. Also implemented proper egglog-based equality checking.


jafioti:
I'm not going to merge this in yet because there are performance issues with egglog simplification. Need to fix those before merge.

jafioti:
ok perf issues should be fixed now

Issue #204, opened 2026-01-08
Argsort kernel op


jafioti:
Similar to #198 , a dedicated op for argsort. See `GraphTensor.argsort` impl for the hlir implementation.

jonahsamost:
Will pick this up!

PR #207, opened 2026-01-11
Native and Cuda Argsort

jonahsamost:
Adds argsort as a primitive op with Native and Cuda implementations.

The cuda kernel uses odd even transposition sort within a single thread block. Shared memory is used for values N <= 4096. More efficient sorting would require some type of global sync.

Referencing #204

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -1536,7 +1536,7 @@ impl Runtime for NativeRuntime {
 }
 
 impl NativeRuntime {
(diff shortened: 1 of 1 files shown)


jafioti:
Argsort should remain an HLIR op, the idea, similar to the MeanReduce, is that the KernelArgsort will match on the HLIR pattern, rather than requiring a dedicated op.

jafioti:
Looks great!

PR #233, opened 2026-02-06
Add GraphTensor::repeat with shape rank validation

MagellaX:
Issue: Closes https://github.com/luminal-ai/luminal/issues/226

### DIFF
--- a/src/frontend/movement.rs
+++ b/src/frontend/movement.rs
@@ -61,6 +61,10 @@ impl GraphTensor {
             input_dims.len(),
             repeats.len()
         );
+        assert!(
+            repeats.iter().all(|r| r.to_usize() != Some(0)),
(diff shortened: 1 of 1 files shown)


MagellaX:
any takes @jafioti

jafioti:
I think it looks correct, but I'm  going to add back in 'z' striding in tensor strides so that this can be done zero cost (not needing a gather op).

MagellaX:
> I think it looks correct, but I'm going to add back in 'z' striding in tensor strides so that this can be done zero cost (not needing a gather op).

 Sounds good to me. I just wanted to get `repeat` in with correct behavior first (incl. rank + zero-repeat checks). let me know for the zero-cost stride version as the follow-up...

jafioti:
repeat was added with z striding

PR #259, opened 2026-03-30
Zero-copy CUDA weight sharing between PyTorch and luminal

tucker-luminal:
## Summary

Eliminates duplicate GPU memory usage for model weights by sharing CUDA device pointers directly between PyTorch and luminal's runtime, instead of serializing weights to disk and re-uploading them.

### Key changes

- **`CudaRuntime::set_device_ptr()`** — wraps external CUDA pointers in `ManuallyDrop<CudaSlice>` so luminal can read PyTorch's GPU buffers without taking ownership (no `cuMemFree` on drop)
- **`CudaRuntime::persist_hlir_node()`** — prevents weight buffers from being consumed after `execute()`, since the external allocator (PyTorch) owns the memory
- **Unified `CompiledGraph::parse_graph()`** — both ONNX and PT2 paths now produce a format-neutral `GraphTranslation` + `WeightData`, then feed into a single shared compilation pipeline (was duplicated before)
- **Python-side weight detection** — identifies weight vs user inputs from FX graph placeholder names (`l_self_`, `l_model_`, `l__self_`), collects device pointers upfront, and passes them to Rust during compilation
- **`CompiledModel`** tracks `_weight_refs` (prevents GC of shared tensors) and `_user_indices` (only sends actual user inputs per forward call)
- **Zero-copy input path** — `set_input_device_ptr()` for CUDA tensors avoids `.tolist()` / numpy conversion overhead; `set_input_from_ptr()` for CPU tensors avoids Python list conversion

### CUDA kernel additions

- **`KernelExp`** — native `expf()` kernel, avoids precision loss from `exp2(x * 1.442695)` truncation
- **`KernelSigmoid`** — fused `1/(1+expf(-x))` instead of 5 separate kernels
- **Kahan summation** in `SumReduce` for numerical stability across 16+ layer models

### Other improvements

- Detailed `[LUMINAL TIMING]` instrumentation in `build_search_space` and `search_single` for profiling compilation and search phases
- PT2 path no longer requires safetensors file — weights loaded via pointers after compilation
- Refactored `pt2_compiled_model.rs` to produce `GraphTranslation`/`WeightData` like the ONNX path, removing ~200 lines of duplicated runtime init code

## Test plan

- [x] `run_tests_cuda.sh` — CUDA ONNX path (test_llama3.py)
- [x] `run_tests_cuda_fx.sh` — CUDA PT2 path (test_hlir_ops.py + test_unary.py + test_llama3.py)
- [x] Verify GPU memory usage is reduced (weights not duplicated)


crates/luminal_cuda_lite/src/kernel/hlir.rs
```diff
@@ -1647,6 +1653,381 @@ extern \"C\" {{
...
+        )
+    }
+
+    fn n_inputs(&self) -> usize {
+        1
+    }
+
+    fn rewrites(&self) -> Vec<Rule> {
+        vec![
+            // Match via the logical Exp op (created by logical.rs Exp rewrite)
```
jafioti: Why is there a logical Exp? Doesn't this take the place of the logical Exp op?

crates/luminal_cuda_lite/src/kernel/hlir.rs
```diff
@@ -1647,6 +1653,381 @@ extern \"C\" {{
...
+        )
+    }
+
+    fn n_inputs(&self) -> usize {
+        1
+    }
+
+    fn rewrites(&self) -> Vec<Rule> {
+        vec![
+            // Match via the logical Sigmoid op (created by logical.rs Sigmoid rewrite)
```
jafioti: Same here, remove the logical sigmoid since we now have a real sigmoid op

crates/luminal_cuda_lite/src/kernel/hlir.rs
```diff
@@ -58,6 +58,8 @@ pub type Ops = (
     KernelSumReduce,
     KernelMaxReduce,
     KernelExp2,
+    KernelExp,
```
jafioti: Can you please put both of these pps in other_ops.rs? hlir.rs should only contain the exact analogous ops to real HLIR ops

crates/luminal_cuda_lite/src/kernel/hlir.rs
```diff
@@ -451,8 +453,12 @@ extern \"C\" {{
         long long iters = {iters};
 
         {dtype} partial = 0;
+        {dtype} comp = 0;   // Kahan compensation
         for (long long i = tid; i < iters; i += THREADS_PER_BLOCK) {{
-            partial += in_data[in_start + {iter_stride_of_i}];
```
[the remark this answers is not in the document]
asglover: @jafioti @tucker-luminal I don't think we can do this by default. I think we need to revert this. Kahan summation is a great low cost way to improve accumulation accuracy, but it's philosophically very similar to upcasting to a higher precision accumulator, or in the opposite direction, turning on TF32. I think the simple kernel sum needs to stay a simple accumulation, and then there should be a toggle-able rewrite to rewrite summations into Kahan summations, then match a Kahan summation kernel onto the Kahan summation pattern. It's more steps but it's the only way we can be consistent and explicit. It's fine as is (it's already merged) but I'll revert it and re treat it in a separate pr.

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -299,124 +108,154 @@ impl CompiledGraph {
...
+                    rt.set_data(node_id, data.clone());
+                }
             }
         }
-
-        // Now finalize (search with profiling, data is available)
-        let cuda_rt = finalize_cuda(context, cuda_rt);
-
-        Ok(cuda_rt)
+        trace!(
```
asglover, line 232: This is too specific and specialized and should get removed in a different pr

asglover: It doesn't look like it works for various dtypes?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -525,6 +364,94 @@ impl CompiledGraph {
         Ok(())
     }
 
+    /// Set input tensor data from a CPU host memory pointer (avoids Python list conversion).
+    /// The pointer must point to contiguous f32 data (from tensor.data_ptr() on a CPU float32 tensor).
```
asglover, line 368: be aware that not all pytorch tensors are guaranteed to be contiguous. There is an API for checking if they are contiguous (and requesting a contiguous version if they are not)

asglover: maybe this is checked elsewhere, I'm not sure

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -525,6 +364,94 @@ impl CompiledGraph {
...
+            PyErr::new::<pyo3::exceptions::PyKeyError, _>(format!("Unknown input tensor: {}", name))
+        })?;
+        let data: Vec<f32> =
+            unsafe { std::slice::from_raw_parts(ptr as *const f32, n_elements).to_vec() };
+        self.runtime.set_data(*node_id, data);
+        Ok(())
+    }
+
+    /// Set input from a CUDA device pointer. Zero-copy on device.
+    /// The pointer must be a valid CUDA device allocation with at least n_bytes bytes.
```
asglover, line 381: similar comment. I feel like weights and inputs should eventually both be pytorch tensors. The device location type (and which device) should be available in the fields for those tensors. I wonder if there's a way to unify the way that we interact with them? no specific action.

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -525,6 +364,94 @@ impl CompiledGraph {
...
+                    "set_input_device_ptr requires CUDA backend",
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    /// Mark an input tensor as persistent (survives execute() calls).
+    /// Call this for weight tensors that should not be consumed after each execution.
+    fn persist_input(&mut self, name: &str) -> PyResult<()> {
```
asglover, line 405: @jafioti @tucker-luminal Having persist input be handled differently per runtime feels strange? Why wouldn't this be a characteristic / specification that comes from the input graph and is respected by both backends?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -525,6 +364,94 @@ impl CompiledGraph {
...
+            RuntimeBackend::Cuda(rt) => rt.persist_hlir_node(_node_id),
+            RuntimeBackend::Native(_) => {} // Native: persist is handled at graph level
+        }
+        Ok(())
+    }
+
+    /// Set a weight tensor from a CUDA device pointer, matching by Input node label.
+    /// Also marks the weight as persistent. For PT2 weights (e.g. "fc1.weight").
+    #[cfg(feature = "cuda")]
+    fn set_weight_device_ptr(
```
asglover, line 420: This is just nitpicking, and does not require action, but can we separate out the tag as "persisting" from the other functions. Do we fundamentally have to treat weights and inputs distinctly. I don't think we should can I think it might create problems later if we do. Could these be treated in a unified way as tensors?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -537,7 +464,7 @@ impl CompiledGraph {
         })
     }
 
-    /// Get output tensor data by name.
+    /// Get output tensor data by name (copies to host).
     fn get_output(&self, name: &str) -> PyResult<Vec<f32>> {
```
asglover, line 468: is the get_f32 thing being made dtype agnostic in your current work?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -547,4 +474,25 @@ impl CompiledGraph {
         })?;
         Ok(self.runtime.get_f32(*node_id))
     }
+
+    /// Copy output tensor data directly to a CUDA device pointer (DtoD).
+    /// Avoids the DtoH + HtoD round-trip of get_output() + .to(device).
+    #[cfg(feature = "cuda")]
```
asglover, line 480: @tucker-luminal @jafioti I don't think we currently can perform this optimization, but we should be able to write directly to the output buffers and be able to eliminate these copies. if you agree in principle, I'll make a linear backlog ticket for this.

crates/luminal_python/run_tests_cuda.sh
```diff
@@ -14,7 +14,7 @@ uv run maturin develop --manifest-path rust/Cargo.toml --features cuda -r
 
 # Run pytest with CUDA backend
 echo "Step 3: Running pytest with CUDA backend..."
-RUST_BACKTRACE=1 LUMINAL_BACKEND=cuda uv run pytest tests/test_hlir_ops.py tests/test_unary.py tests/test_llama3.py -m "not slow" -v
```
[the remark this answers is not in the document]
asglover: are now fast enough?

crates/luminal_python/rust/src/compiled_graph.rs
tucker-luminal: Yup :) all of the f32 stuff was done for making the first pass on all of this easier, addressing it now

crates/luminal_python/rust/src/compiled_graph.rs
jafioti: currently you can see in the llama rust example that we can basically re-use input buffers for output buffers (with ScatterNoCopy) because the input is consumed by execution. Is this different than that?

asglover: I think this is a different issue / idea. I think what this is saying is that when finished, copy the tensor to the waiting pytorch buffer. Eventually when we have buffer semantics, we should be able to just have the final operation write directly to the waiting pytorch buffer instead of needing this explicitly copy operation.

crates/luminal_cuda_lite/src/kernel/hlir.rs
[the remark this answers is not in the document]
jafioti: Yeah that makes sense to me. We should basically handle the precision changing api (incl fast math / tf32) at the same time as focusing more on lower precisions.

crates/luminal_python/rust/src/compiled_graph.rs
jafioti: Agreed, sorry I rushed this review. There is already a .persist() function on graphtensor, so I don't think this is needed?

crates/luminal_python/rust/src/compiled_graph.rs
jafioti: Right, again with tensors already persisting and in the correct locations from pytorch, we should not need anything specific in the runtime

crates/luminal_python/rust/src/compiled_graph.rs
jafioti: This should be handled in the stabilized runtime interface PR, which I think is the logical next step here

crates/luminal_python/rust/src/compiled_graph.rs
jafioti: Yeah having a buffer provided for outputs makes sense to me. I think it needs to be a field on the Output node though which would then prevent the ScatterNoCopy type optimizations from being done

PR #265, opened 2026-04-10
build MoE routing support in luminal_python

alityb:
Before this change, a tiny exported routing graph was not supported: translation would fail on `aten.argsort.default`, and compiled-model output handling did not preserve routing int/bool artifacts correctly. 

Changes:

On the luminal_python side, this adds the following ops: argsort, topk, scatter.value, masked_fill.Scalar, bitwise_not, floor_divide, sign. 

It also fixes mixed-dtype routing math and updates compiled-model dtype handling so routing indices and masks stay as int and bool instead of being forced through f32. 

On the CUDA side, I only restored the backend behavior needed for this slice. Grouped cuBLASLt host ops are now gated off on machines that do not have `cublasLtGroupedMatrixLayoutCreate`, instead of being picked during search and failing later. This is the only backend-side change I kept because it directly affected the routing validation path.

Proof:
The main proof is kept small and active:

- test_argsort_stable_duplicates
- test_tiny_moe_routing

These cover stable ordering, regrouping, gather/scatter, masking, and mixed int / bool / float routing outputs.

Validation

- cargo fmt
- cargo clippy -p luminal_python --all-targets -- -D warnings
- uvx ruff check .
- uv run pytest tests/test_hlir_ops.py -q
   - 195 passed
- uv run pytest tests/test_unary.py -q
   - 16 passed
- uv run pytest tests/test_hlir_ops.py -q -k "test_gather_embedding or test_gather_2d_axis0 or test_argsort_stable_duplicates or test_tiny_moe_routing or test_topk_values or test_topk_indices"
    - 6 passed
- LUMINAL_BACKEND=cuda uv run pytest tests/test_hlir_ops.py -q -k "test_gather_embedding ortest_gather_2d_axis0 or test_argsort_stable_duplicates or test_tiny_moe_routing or test_topk_values or test_topk_indices"
    - 6 passed
- LUMINAL_BACKEND=cuda uv run pytest tests/test_hlir_ops.py -q
    - 193 passed, 2 failed (both Conv which are known failures right now)


tucker-luminal:
This looks great! Running our larger tests while I review

tucker-luminal:
Since this is a test that is harder for you to look at, I can do some investigation on our end and help you out

tucker-luminal:
Hey! I took a look at the OOM in the qwen3_moe modal
                                                                                                                     
  The cublaslt_grouped_layout_supported() check in luminal_cuda_lite is gating off too much. When
  cublasLtGroupedMatrixLayoutCreate isn't found via dlopen, it disables all CuBlasLt matmul rewrite rules and the    
  entire fused GLUMoE op. Without those, every matmul falls back to unfused Mul + SumReduce (huge intermediates)  and   the MoE path materializes all top-k experts' weights at once instead of processing them sequentially in the fused  
  op. That's what pushes the 30B model past 80GB.       

  The good news is that neither CuBlasLt nor GLUMoE actually use the grouped layout API at runtime — they both use   
  standard cublasLtMatrixLayoutCreate + cublasLtMatmul (with strided-batch attributes for batched cases). So the
  check isn't needed for either of them. The try_create_cublaslt() fallible init you added already handles the case  
  where cuBLASLt is genuinely unavailable at execution time, which is the right place for that safety net.

  The fix is just removing the three-line guard from CuBlasLt::rewrites() and GLUMoE::early_rewrites(), cleaning up  
  the unused imports, and removing the now-dead cublaslt_grouped_layout_supported() function from lib.rs. I verified
  all native and CUDA tests still pass with this change.

alityb:
> Hey! I took a look at the OOM in the qwen3_moe modal
> 
> The cublaslt_grouped_layout_supported() check in luminal_cuda_lite is gating off too much. When cublasLtGroupedMatrixLayoutCreate isn't found via dlopen, it disables all CuBlasLt matmul rewrite rules and the entire fused GLUMoE op. Without those, every matmul falls back to unfused Mul + SumReduce (huge intermediates) and the MoE path materializes all top-k experts' weights at once instead of processing them sequentially in the fused op. That's what pushes the 30B model past 80GB.
> 
> The good news is that neither CuBlasLt nor GLUMoE actually use the grouped layout API at runtime — they both use standard cublasLtMatrixLayoutCreate + cublasLtMatmul (with strided-batch attributes for batched cases). So the check isn't needed for either of them. The try_create_cublaslt() fallible init you added already handles the case where cuBLASLt is genuinely unavailable at execution time, which is the right place for that safety net.
> 
> The fix is just removing the three-line guard from CuBlasLt::rewrites() and GLUMoE::early_rewrites(), cleaning up the unused imports, and removing the now-dead cublaslt_grouped_layout_supported() function from lib.rs. I verified all native and CUDA tests still pass with this change.

Thanks a lot for the help, I was pretty stuck on trying to fix it while testing locally but that's a cool subtlety of CuBlasLt API. Done and it seemed to work, should be all good now :)

tucker-luminal:
Some changes got pulled into main, if you could handle them this is good to merge

PR #274, opened 2026-04-21
Elementwise fusion for adjacent unary kernels in cuda_lite

matthewjgunton:
## Summary
- New `KernelFusedElementwise` LLIR op in `luminal_cuda_lite` that collapses a chain of pure-elementwise unary kernels (Sin/Sqrt/Exp2/Log2/Recip) into a single CUDA kernel.
- Egglog rules: 25 pair rules (unary+unary → Fused[2]) + length-explicit extend rules up to `MAX_FUSION_DEPTH = 8` (Fused[K] + unary → Fused[K+1]). A recursive list-append helper was tried first and exploded the egraph, so the final design is deliberately bounded.
- Fires only when producer out-strides == consumer in-strides (non-linear `?strides` match in the rule), so permutes/reductions break fusion as expected.
- `CudaRuntime::print_kernel_summary()` reports which fused op sequences survived extraction. On Llama-3-8B it shows 33× `[Sqrt, Recip]` + 4× `[Exp2, Recip]` from RMSNorm and rotary-style patterns.

## Test plan
- [x] `cargo test -p luminal_cuda_lite fusion::` — 7/7 pass (pair, 3-op, 4-op structural reachability; stride/reduction negative cases; end-to-end numerical for 2-op and 3-op chains)
- [x] `cargo build -p llama --release` — clean
- [x] Llama-3-8B end-to-end on H100: coherent generation, TTFT ≈ 820ms, TPOT ≈ 39ms, kernel summary confirms Fused kernels are compiled and selected
- [x] Sanity-checked on Qwen3 dense

### DIFF
--- a/crates/luminal_cuda_lite/src/kernel/other_ops.rs
+++ b/crates/luminal_cuda_lite/src/kernel/other_ops.rs
@@ -1809,6 +1809,12 @@ pub struct KernelFusedElementwise {
     dtype: DType,
 }
 
+impl KernelFusedElementwise {
+    pub fn ops(&self) -> &[UnaryFn] {
+        &self.ops
+    }
+}
+
 impl EgglogOp for KernelFusedElementwise {
     fn sort(&self) -> SortDef {
         sort(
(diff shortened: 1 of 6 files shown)


matthewjgunton:
### Before/after benchmark — H100 PCIe, single run, 500-token generation

| model      | metric      | main      | this PR   | Δ        |
|------------|-------------|-----------|-----------|----------|
| Llama-3-8B | TTFT        | 829.95 ms | 826.51 ms | −0.4 %   |
| Llama-3-8B | TPOT        | 39.36 ms  | 39.07 ms  | −0.7 %   |
| Llama-3-8B | search time | 20.7 s    | 107.9 s   | +5.2×    |
| Qwen3-4B   | TTFT        | 432.05 ms | 431.20 ms | −0.2 %   |
| Qwen3-4B   | TPOT        | 37.49 ms  | 38.04 ms  | +1.5 %   |
| Qwen3-4B   | search time | 18.8 s    | 84.3 s    | +4.5×    |

Fusion fires heavily — 92 fused kernels on Llama (61× `[Sqrt, Recip]` + 31× `[Exp2, Recip]`) and 142 on Qwen (141× `[Sqrt, Recip]` + 1× `[Exp2, Recip]`) — but end-to-end wallclock is within noise on a single run. Expected: the unary chains we collapse here (RMSNorm's `Sqrt→Recip`, rotary's `Exp2→Recip`) aren't on the critical path, so saving ~92 launches doesn't surface against the matmul/attention cost.

The real win lands once we extend this to **binary elementwise fusion** (pulling the surrounding `Mul`/`Add` into the same kernel so the RMSNorm body is one launch instead of a chain). This PR is the scaffolding for that — the LLIR op, the stride-matching guard, and the egglog rule shape all carry over.

The +5× search-time cost is worth flagging separately; if it bites in CI we can trim the rule set or cap search depth.

jafioti (approved):

Issue #194, opened 2025-12-30
Llama example does not compile for CPU


Dzuchun:
At the moment, Llama example fails to compile for CPU. In fact, features described in the README are not defined at all.

I was able to compile and launch the example, by adding `load_safetensors` to `NativeRuntime` (using one from `CudaRuntime` as a reference), but the example panics at runtime.

If my understanding is correct, `NativeRuntime` ops do not implement some of the egg functions used by Llama model (namely, `RowRope` is the one causing the "UnboundFunction" error, but I think there are a bunch of them)

Considering this,
- let me know, if Llama is planned to be supported for `NativeRuntime`
- if it is, how stable are APIs relevant this problem at the moment?
- if they are stable, I'd be happy to get directions on the intended fix

I sincerely believe in this project's idea, and would like to contribute

Just in case, here's the panic message I get:
```
called `Result::unwrap()` on an `Err` value: TypeError(UnboundFunction("RowRope", In 651:11-92: (RowRope (ECons (MVar "s") (ENil)) t311 (ECons (MNum 4096) (ENil)) (MNum 4096) t1)))
```
`RUST_BACKTRACE` points to
```
   4: luminal::graph::Graph::build_search_space
             at [...]/src/graph.rs:189:14
```

jafioti:
@Dzuchun Yep the general README at the project's root has changed to reflect that there are no feature flags for llama example currently. technically we can add a native rope op and native attention op, but realistically the native backend is a definitional backend (it's mostly a semantics spec) meaning it's supposed to be simple and correct, not fast.

bringing back the CPU backend is definitely something on the roadmap though, and if you'd like to work on it i'd be happy to give guidance. essentially it would be similar to the cuda backend in that it would define ops that the HLIR ops would be rewritten to that actually prioritize speed, and it would use features like simd / avx if available. a good start would be to directly copy the HLIR ops and use the nativeop implementations of them first, and have a matmul op that uses the `matrixmultiply` crate. then more ops can be defined from there / we can look at automatic vectorization / fusion.

viralbhadeshiya:
@jafioti is anyone working on this? otherwise I would like to start.

jafioti:
I'm not sure if @Dzuchun is, but it seems like not so go for it! Please comment here or in the discord if you have questions. My advice I gave before still applies.

viralbhadeshiya:
Okay I'll start working on this. Will ask question on discord. Thanks

PR #283, opened 2026-04-29
Feat/luminal cpu

viralbhadeshiya:
**This PR adds `luminal_cpu` - CPU backend for Luminal**
Issue #194

## What this adds

A new `crates/luminal_cpu` crate that implements a CPU backend for Luminal. This is not complete Luminal_CPU implementations but it's basic start.

**Structure**:
- `src/runtime.rs` — `CpuRuntime` implementing the `Runtime` trait: `load_llir`, `execute`, `allocate_intermediate_buffers`, `get_f32`, `set_data`
- `src/kernel/mod.rs` — `CpuKernelOp` trait and `CpuOps` type alias
- `src/kernel/ops.rs` — all 14 primitive ops with egglog rewrite rules and strided CPU `process()` implementations
- `src/kernel/matmul.rs` — `CpuMatmulDescriptor` pattern matcher and `CpuMatmul` backed by the `matrixmultiply` crate
- `src/tests.rs` — correctness tests for every op, using plain Rust reference implementations
- `examples/cpu` — runnable demo (matmul, elementwise, softmax, two-layer MLP)

## Key design decisions

- Everything pattern-matches in egglog. There is no Rust-side post-pass that walks the extracted LLIR to fuse or select ops. Matmul fusion lives in egglog as an Op (Sum) ↔ Op (CpuMatmul) union with stride-pattern facts, scheduled in the matmul_backend ruleset after kernel_lower. The runtime just executes whatever LLIR egglog hands it.

- Strided indexing everywhere. Every op's process() evaluates a_strides, b_strides, and in_stride from the eggraph to compute flat buffer indices, so broadcasting (stride=0) and non-contiguous layouts just work without special-casing.

- 'z' substitution. Luminal uses 'z' as a symbolic element-stride variable in all expressions. The CPU backend substitutes z=1 before evaluating any expression, since CPU buffers are flat f32 arrays indexed by element count.

## What's not in this PR (follow-up)

- Batched matmul (currently falls back to elementwise Mul + SumReduce)
- SIMD/AVX intrinsics beyond what `matrixmultiply` provides automatically
- Additional HLIR ops (RoPE, attention) — the owner noted these as future work

## How to test

**Run the test suite:**
```bash
cargo test -p luminal_cpu
```

**Run the demo example:**
```bash
cd examples/cpu
cargo run --release
```


jafioti:
We're currently making a big-ish change that should help to solidify some of the semantics around "consumed" inputs, so I'm going to hold off on a full review of this until those are merged. Probably should be done in the next week or two

PR #290, opened 2026-05-02
Fix MetalGather output dtype inference

spinlocked:
MetalGather was using the default kernel dtype inference, which takes the first input dtype. For gather, the first input is the Int index tensor and the second input is the gathered data tensor, so F32 gathers were compiled with Int outputs.

Infer the output dtype from the data input instead.


jafioti:
Good catch!

PR #293, opened 2026-05-03
Fix Metal index arithmetic and non-contiguous gather lowering

spinlocked:
Metal binary kernels were reading Int inputs through float conversion, which could lose precision for large computed indices. Keep Add, Mul, and Mod in integer space when the output dtype is Int, and use the integer `%` operator for Int modulo.

MetalGather also lowered gathered data offsets using the output/index shape instead of the source data shape. Thread data_shape through the MetalGather egglog op and use it with data_strides when computing the final data index, so gathers from transposed or otherwise non-contiguous tensors address the right elements.


jafioti:
excellent catch, good fix

PR #297, opened 2026-05-05
Add Metal support for Qwen3-4B generation

spinlocked:
Extend MetalRuntime with the runtime APIs needed for loading safetensors, managing persistent KV-cache buffers, round-tripping output buffers back into inputs, and reading logits during autoregressive decoding.

Update the Qwen example to support both CUDA and Metal through mutually exclusive CUDA and Metal feature flags.

### DIFF
--- a/ci/modal_example.py
+++ b/ci/modal_example.py
@@ -45,6 +45,10 @@ EXPECTED_OUTPUT = {
     ],
 }
 
+EXAMPLE_CARGO_ARGS = {
+    "qwen": ["--features", "cuda"],
+}
+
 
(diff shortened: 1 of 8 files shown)


jafioti:
@spinlocked  This is great work! Is it possible instead of having two seperate examples with a third common crate, to just have one example with the cuda and metal stuff gated with mutually exclusive feature flags, like "metal" and "cuda"?

spinlocked:
@jafioti Yes, that does seem better. I’ve updated the changes to do that.

jafioti:
@spinlocked did you want to add more to this PR or want me to review it now?

spinlocked:
@jafioti Should be ready for review

jafioti:
looks great!

PR #298, opened 2026-05-05
luminal_python: gather-then-matmul lowering for grouped_mm

tucker-luminal:
translate_grouped_mm was casting the full [G, K, N] expert weight tensor to F32 before a broadcast batched matmul, producing ~2.1 GB of intermediate buffers per layer on Qwen3-30B-A3B. Across 48 MoE layers this OOM'd the search profiler at runtime.rs:711 (alloc_zeros), failing every python_luminal qwen3-moe bench run for the past ~2 weeks.

Switch to the gather-first pattern that examples/qwen3_moe uses: compute expert_id from offs, gather only the [S, K, N] active slice, then matmul. The shape mirrors what glumoe_rewrite.egg matches, and the gather is 16x smaller at prefill
(S = num_tokens * top_k = 8 vs G = 128).

Two refinements baked in vs the broadcast-and-mask version:

1. Stay in Int for the entire expert_id computation. arange and offs are already Int; ge → Bool → cast(Int) → sum → minimum handles the clamp without four F32 round-trips. Same value as HF MoE's `expert_ids.clamp(0, num_experts-1)` for invalid expert IDs from EP, AND protects search-time profiling: dummy-1 input bytes give offs=[1,…,1], pushing the raw count to G for any token with index ≥ 1, which would OOB the gather without the clamp.

2. Drop the cast(F32) on input and on the gathered weight. The broadcast-and-mask version needed F32 because it casted the mask to F32; gather-then-matmul has no such requirement, and casting `[S, K, N]` to F32 doubled the gather scratch (~100 MB → ~200 MB per layer for Qwen3-30B-A3B prefill). Matmul rewrites (cuBLASLt etc.) handle bf16 input with F32 accumulator internally — no precision loss in practice.

Verification:
- tests/test_hlir_ops.py::test_grouped_mm_fallback{,_routing_invariance} pass.
- Synthetic g=128, s=8, k=2048, n=1536 bf16 test: max-abs-diff 1.56e-02 (within bf16 accumulation tolerance; expected to drop to F32-accurate once the cuBLASLt rewrite fires at higher search budgets).

Result: original OOM-in-search is gone. With --search-iters 1 the full Qwen3-30B-A3B bench end-to-ends (TTFT ~9.4s).


jafioti:
Awesome! The cuda unit test failure i think is a timeout issue I fixed on main, should be fixed for the future.

PR #307, opened 2026-05-11
capacity should be used instead of len for Vec::from_raw_parts

39ali:
`Vec::from_raw_parts` is called with `capacity = n (= len / size_of::<T>())`, but the original allocation may have `capacity > len`. On drop, the allocator gets the wrong size, which is UB


jafioti:
Good catch!

PR #314, opened 2026-05-12
Add dynamic KV-cache llama chat server

tucker-luminal:
An update to how we handle dynamic dimensions in python


crates/luminal_python/rust/src/translator/tensor.rs
```diff
@@ -274,8 +275,24 @@ impl<'a> Translator<'a> {
...
-        // Gather → [S, K, N], preserves weight's native dtype (bf16 stays bf16).
+        // Gather → [S, K, N], then normalize both operands to the op's declared
+        // output dtype before matmul. On real Qwen3-MoE bf16 checkpoints the FX
+        // graph inserts casts on the activation path, and relying on the input
+        // tensor's translated dtype can leave us with mixed F32/Bf16 operands
+        // by the time matmul expands into elementwise Mul. Using the PT2 output
+        // metadata keeps the matmul dtype aligned with the exported contract
+        // without upcasting the full expert weight bank.
         let weight_gathered = weight.gather(flat_idx);
+        let input = if input.dtype == out_dtype {
```
asglover: Why can't we just cast to the type we want? And in the situations where it's the same it's a no-op?

crates/luminal_python/rust/src/translator/tensor.rs
```diff
@@ -287,7 +304,11 @@ impl<'a> Translator<'a> {
         // (cuBLASLt etc.) handle bf16 input with F32 accumulator internally.
         let result = input.unsqueeze(1).matmul(weight_gathered).squeeze(1);
 
-        Ok(result.cast(input.dtype))
+        Ok(if result.dtype == out_dtype {
```
asglover: same

crates/luminal_python/rust/src/translator/binary.rs
```diff
@@ -54,4 +67,18 @@ impl<'a> Translator<'a> {
...
     }
+
+    pub(crate) fn apply_symbolic_scalar_op(
+        &mut self,
+        a: GraphTensor,
+        val: Expression,
+        op: BinaryOp,
+    ) -> GraphTensor {
+        match op {
+            BinaryOp::Add => a + val,
```
asglover, line 118: Is this sufficient? Does this cover all cases?

crates/luminal_python/rust/src/translator/mod.rs
```diff
@@ -339,40 +336,49 @@ impl<'a> Translator<'a> {
...
-            if let Some(hint) = e.as_expr.hint.as_ref().and_then(|h| h.as_int()) {
-                return Some(Expression::from(hint as usize));
-            }
+            return self.resolve_expr_value(&e.as_expr);
         }
         None
     }
+
+    pub(crate) fn resolve_expr_str(&self, expr_str: &str) -> Option<Expression> {
+        parse_sympy_expr(expr_str, &self.sym_map.sym_to_char).or_else(|| {
```
asglover: Is there a way we could just call sympy bindings to resolve the sympy expressions directly? or is translating to our system the correct call. I don't know. Maybe translating to our system in the move? If we're bringing into our system we need to make sure we have parity and not a ton of bloaty overhead. Once we get the concrete inputs, should we be able to resolve all sympy expressions right? Or do data dependent operations introduce dynamic dims into graphs?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -127,6 +130,8 @@ pub struct CompiledGraph {
     pub output_dtypes: Vec<DType>,
     pub input_shape_exprs: Vec<Vec<Expression>>,
     pub dim_param_map: DimParamMap,
+    static_host_inputs: StaticHostInputs,
```
asglover: What are these for? What does static mean? Why do we need to special case for static inputs?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -151,17 +156,21 @@ impl CompiledGraph {
             input_shape_exprs,
             dim_param_map,
         } = translation;
+        let WeightData {
```
asglover, line 161: similarly, why do we need to special case on weights? What does treating some inputs as weights enable?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -151,17 +156,21 @@ impl CompiledGraph {
...
+            weights,
+            tensor_sizes,
+            device_ptrs,
+        } = weight_data;
 
         // Build compile args from WeightData (convert TypedData -> raw bytes + dtype)
```
asglover: Should we actually be passing. just the pointer? What happens when a "strided"  / non-contiguous tensor is passed and we need more than just the pointer?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -380,7 +426,8 @@ impl CompiledGraph {
         Ok(())
     }
 
-    /// Set a weight from a device pointer (e.g. "fc1.weight"). Zero-copy on device.
+    /// Register a weight from a device pointer (e.g. "fc1.weight"). Zero-copy on device.
+    /// The registration persists across future run() calls.
```
asglover: Should that be the case? Are we sure that the base pointer that backs a tensor will be the same between calls? I'm not sure? Would that trigger a recompile? I'm not sure if it should?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -537,3 +587,50 @@ impl CompiledGraph {
...
+        }
+
+        let static_host_inputs: Vec<(NodeIndex, TypedData)> = self
+            .static_host_inputs
+            .iter()
+            .map(|(&node_id, typed)| (node_id, typed.clone()))
+            .collect();
+        for (node_id, typed) in static_host_inputs {
+            self.runtime
+                .set_data_bytes(node_id, typed.bytes, typed.dtype);
```
asglover: why are we still device vs. host specialized?

crates/luminal_python/rust/src/pt2_compiled_model.rs
```diff
@@ -45,147 +46,14 @@ fn resolve_dim_sizes(
                             .hint
                             .as_ref()
                             .and_then(|h| h.as_int())
-                            .map(|h| Expression::from(h as usize))
+                            .map(Expression::from)
```
asglover, line 49: what's the difference in how we and them represent dimensions and how pytorch does it at the type level?

crates/luminal_python/rust/src/pt2_expr.rs
```diff
@@ -0,0 +1,194 @@
...
+    match head {
+        "Symbol" => {
+            let name = extract_first_quoted(body)?;
+            sym_to_char.get(&name).map(|c| Expression::from(*c))
+        }
+        "Integer" | "Number" => {
+            let value = body.trim().parse::<i64>().ok()?;
+            Some(Expression::from(value))
+        }
+        "NegativeOne" => Some(Expression::from(-1i64)),
```
asglover: what? why is this special cased?

crates/luminal_python/rust/src/pt2_expr.rs
```diff
@@ -0,0 +1,194 @@
...
+        "Zero" => Some(Expression::from(0i64)),
```
asglover: same with these?

crates/luminal_python/rust/src/pt2_expr.rs
```diff
@@ -0,0 +1,194 @@
...
+fn looks_like_kwarg(part: &str) -> bool {
+    if let Some(eq) = part.find('=') {
+        let key = part[..eq].trim();
+        return !key.is_empty() && key.chars().all(|c| c == '_' || c.is_ascii_alphanumeric());
+    }
+    false
+}
+
+#[cfg(test)]
+mod tests {
```
asglover: Do we know what we need to test against for sympy parsing? It's good to have tests, but why these couple tests only? Do these cover the main features?

crates/luminal_python/tests/test_dynamic_shapes.py
```diff
@@ -34,6 +37,64 @@ def wrapper(gm, example_inputs):
...
+def _first_trace_dynamic_shapes(capture_holder):
+    from luminal.pt2 import (
+        _build_dynamic_shapes_from_gm,
+        _reinternalize_lifted_params,
+        _strip_symint_placeholders,
+    )
+
+    gm = copy.deepcopy(capture_holder["gm"]).eval()
+    example_inputs = capture_holder["example_inputs"]
+    gm, user_inputs, _, _ = _reinternalize_lifted_params(gm, example_inputs)
```
asglover, line 77: I don't understand what's going on with these tests. Why is this necessary / what is this testing? We can go through this if you'd like, I don't understand the pattern.

crates/luminal_python/tests/test_static_weight_replay.py
```diff
@@ -0,0 +1,175 @@
...
+            )
+    finally:
+        tmpdir.cleanup()
+
+
+@pytest.mark.skipif(
+    not torch.cuda.is_available() or _cuda_lite_factory_capsule is None,
+    reason="CUDA-only — exercises persistent zero-copy device weight replay",
+)
+def test_process_pt2_set_weight_device_ptr_persists_across_runs():
```
asglover: The name of this test is spooky. Why would the weight ptr persist across runs? Should we assume that?

crates/luminal_python/LessonsLearned.md
```diff
@@ -867,3 +867,39 @@ Also: search-time dummy-1 inputs are not the same shape as runtime inputs. Anyth
...
+   full `[batch, seq, vocab]` logits tensor even though generation only uses the last token, so
+   each step also paid for an unnecessary sequence-wide LM head projection and a larger output
+   buffer.
+3. **Why it was tricky**: the slowdown looked like “performance drifting over time,” which can
+   be mistaken for memory fragmentation or a leak. The server stayed healthy and GPU memory was
+   stable, so the real issue was algorithmic. The obvious optimization — slice to the last token
+   inside the compiled graph — ran straight into a separate backend/compiler bug in the current
+   dynamic CUDA path.
+4. **Fix in this session**:
+   - Kept the dynamic path for fast short responses.
```
asglover: This seems like an incorrect way of solving the problem? Why would we cap the number of tokens a new token can attend to artificially? why is this in lessons learned?

asglover (dismissed):
Sorry for these pedantic comments, hopefully some are helpful

crates/luminal_python/rust/src/translator/mod.rs
tucker-luminal: See what happens when you pass in a Min and Max dimension

crates/luminal_python/rust/src/translator/tensor.rs
tucker-luminal: Where did this come from, can we do what Austin wants here

crates/luminal_python/rust/src/translator/tensor.rs
tucker-luminal: same as above

crates/luminal_python/rust/src/compiled_graph.rs
tucker-luminal: Take a look at why .presist does not work

crates/luminal_python/rust/src/compiled_graph.rs
tucker-luminal: same as above

crates/luminal_python/rust/src/compiled_graph.rs
tucker-luminal: Write a test for this, figure out the total information coming over if possible

PR #323, opened 2026-05-19
Dtype i64 f64 first class

tucker-luminal:
## First-class I64 / F64 in luminal IR + strict FFI dtype boundary

Today luminal collapses every PT2 integer dtype to `DType::Int` (i32) and
`float64` to `DType::F32` at the FFI boundary, then casts back at the
Python wrapper. Values outside the i32 / f32 range (`2**40`,
`1.0000000000000002`) lose information before the kernel ever runs.

This PR makes I64 and F64 first-class end-to-end:

- **IR**: `DType::I64` added (debug-formatted as `"Int64"` to dodge
  egglog's built-in `I64` primitive sort). `NativeData::I64` /
  `NativeData::F64` + accessors. `Cast` covers the full new Cartesian
  product — cast to `Int` from `I64` saturates to match
  `tensor.to(torch.int32)`.
- **CPU**: `Add`, `Mul`, `Mod`, `Gather`, `Scatter`, `SumReduce`,
  `MaxReduce` handle I64 / F64 directly. Unary transcendentals still
  bridge through f32 in v1; reaching the kernel with `I64`/`F64`
  panics with a pointer to the missing bridge.
- **CUDA**: `cuda_dtype` extended (`"long long"` for I64). Output read
  paths (`get_output_i64` / `get_output_f64`) added; existing buffers
  in narrower dtypes are widened on the read side. I64 is now in the
  scatter elem-size tables.
- **PT2 boundary**: `torch_dtype_int_to_luminal` returns `I64` / `F64`
  for PyTorch codes 5 / 8. `TypedData::from_pytorch_bytes` and
  `pt2_compiled_model::bytes_to_typed` preserve raw bytes for both.
  `luminal_dtype_to_pt2_code` round-trips `I64` to code 5.
- **Python wrapper**: `torch.int64` / `torch.float64` outputs route
  through the new `get_output_i64` / `get_output_f64` paths — no more
  i32-buffer-then-`.to(int64)` cast-back layer. CUDA pre-allocates the
  f64 tensor but skips the zero-copy device-ptr handoff (kernels still
  write f32 bytes there in v1; Phase F will land real f64 codegen).
- **Boundary contract**: `model(x)` now requires
  `x.dtype == model.input_dtypes[i]` — the previous behavior cast
  silently at every call, hiding precision bugs and burning cycles on
  per-call allocation+copy. Convert at the call site with
  `x.to(model.input_dtypes[i])` if you need a different dtype.

## Tests

`crates/luminal_python/tests/test_dtype_boundary.py` covers the new
behavior. `xfail`s removed from `int64_outside_i32_range` and
`float64_precision_sensitive` — both pass on CPU end-to-end now. CUDA
parity for elementwise i64 / f64 kernels lands in Phase F.

Verified by `./run_all_tests.sh` (native + CUDA non-slow):
`test_hlir_ops.py + test_unary.py + test_dtype_boundary.py` clean on
both backends.


asglover:
" Unary transcendentals still
bridge through f32 in v1; " 

This is probably worth fixing. It should error or natively support the wider dtype. No implicit casting should be allowed.

asglover:
" Output read
paths (get_output_i64 / get_output_f64) added; existing buffers
in narrower dtypes are widened on the read side" 

Casting needs to be explicitly in the graph, we should never insert an implicit cast

crates/luminal_cuda_lite/src/runtime.rs
```diff
@@ -651,6 +651,78 @@ impl CudaRuntime {
             .collect_vec()
     }
 
+    /// Read a tensor's output bytes back as i64.
+    ///
+    /// The actual buffer dtype may not be I64 — some ops (`Argsort`,
+    /// `TopK`, ...) emit i32 indices today even when the user-visible
+    /// output dtype is `torch.int64`. We dispatch on the runtime's
+    /// per-buffer dtype tag and widen from i32 / i16 / i8 / bool when
+    /// needed, mirroring the CPU `NativeData::i64()` accessor.
```
asglover: we should never "insert a widen when needed" unless a cast is in the graph.

crates/luminal_nn/src/attention.rs
```diff
@@ -275,14 +275,14 @@ mod tests {
         let mut rt = cx.search(NativeRuntime::default(), 1);
 
         // Q = [1, 0, 1, 0] → head0=[1,0], head1=[1,0]
-        rt.set_data(q.id, vec![1., 0., 1., 0.]);
+        rt.set_data(q.id, vec![1.0f32, 0., 1., 0.]);
```
asglover: is there some way that you can put the type in the vec<type> form instead of annotating the data? Just for readability / style?

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -81,6 +81,7 @@ fn luminal_dtype_to_pt2_code(dtype: DType) -> u32 {
         DType::I8 => 2,
         DType::I16 => 3,
         DType::Int => 4, // i32
+        DType::I64 => 5,
```
asglover: This is a pre-existing sin, but is there a way we can have a rust enum for Pytorch's dtypes? It can be bindgen'd so that it will remain in sync, but that might help with readability / help understand if something janky is being done

crates/luminal_python/rust/src/compiled_graph.rs
```diff
@@ -512,6 +513,36 @@ impl CompiledGraph {
         Ok(self.runtime.get_output_i32(*node_id))
     }
 
+    /// Get output tensor data by name as i64 (copies to host).
+    ///
+    /// Used for `torch.int64` outputs. Reads the native I64 buffer when the
+    /// IR computed in I64 (preserving values outside the i32 range); widens
+    /// i32 / bool when the producer chose a narrower dtype.
```
asglover: It should only be able to get an i64 output if that's what in the graph. If something narrower is in the graph, it needs to be extracted in that size and then up cast externally. we should not do implicit casting.

crates/luminal_python/rust/src/pt2_compiled_model.rs
```diff
@@ -388,26 +388,10 @@ fn bytes_to_typed(bytes: &[u8], dtype: u32) -> TypedData {
...
-            let f32s: Vec<f32> = bytes
-                .chunks_exact(8)
-                .map(|b| {
-                    f64::from_le_bytes([b[0], b[1], b[2], b[3], b[4], b[5], b[6], b[7]]) as f32
-                })
-                .collect();
-            TypedData::from_f32_vec(f32s)
-        }
+        // i64 / f64 are first-class — preserve raw bytes so values outside
+        // the i32 / f32 representable range round-trip through the IR.
```
asglover: remove the comment

crates/luminal_python/rust/src/pt2_util.rs
```diff
@@ -199,15 +199,22 @@ pub fn resolve_neg1_dim_exprs(
...
 /// Map torch dtype integer (PT2 format) to luminal DType.
-/// PT2 numbering: 1=uint8, 2=int8, 3=int16, 4=int32, 5=int64, 6=float16, 7=float32, 8=float64, 12=bool, 13=bfloat16
+/// PT2 numbering: 1=uint8, 2=int8, 3=int16, 4=int32, 5=int64, 6=float16, 7=float32, 8=float64, 12=bool, 13=bfloat16.
+///
+/// `int64`/`float64` are first-class in the IR (`DType::I64`, `DType::F64`)
+/// so values outside the i32 / f32 representable ranges survive round-trip
+/// through luminal arithmetic. Narrower integer widths still collapse to
+/// `DType::Int`; the user-visible dtype is recovered at the Python boundary
+/// via the PT2 dtype-code metadata kept alongside.
 pub fn torch_dtype_int_to_luminal(dtype: u32) -> DType {
```
asglover, line 206: same comment about creating an enum for PT types

crates/luminal_python/rust/src/typed_data.rs
```diff
@@ -152,8 +152,8 @@ impl TypedData {
     /// Convert raw bytes from a PyTorch tensor (identified by PT2 dtype code) to TypedData
     /// in luminal's native format. Handles widening/narrowing conversions for types where
     /// PyTorch's byte layout differs from luminal's:
-    /// - i64 → i32, f64 → f32 (luminal has no 64-bit types)
-    /// - i16 → i32, u8 → i32, i8 → i32 (luminal maps all integer types to i32 for PT2)
+    /// - i64 / f64 preserved as `DType::I64` / `DType::F64` (first-class IR types)
+    /// - i16 → i32, u8 → i32, i8 → i32 (luminal maps narrower integers to i32)
```
asglover: This bit: 

  /// - i16 → i32, u8 → i32, i8 → i32 (luminal maps narrower integers to i32)

scares me, I don't know why we need to do this? why we cannot respect the dtypes and not upcast them.

crates/luminal_python/rust/src/typed_data.rs
```diff
@@ -162,26 +162,10 @@ impl TypedData {
...
-            8 => {
-                let f32s: Vec<f32> = bytes
-                    .chunks_exact(8)
-                    .map(|b| {
-                        f64::from_le_bytes([b[0], b[1], b[2], b[3], b[4], b[5], b[6], b[7]]) as f32
-                    })
-                    .collect();
-                Self::from_f32_vec(f32s)
-            }
+            // i64 / f64 — first-class in the IR; preserve bits so values
```
asglover: same as before

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -8,6 +8,20 @@
 from .dtype_util import torch_dtype_code as _torch_dtype_code
 
 
+class DTypeBoundaryError(TypeError):
```
asglover, line 11: Nice!

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -116,7 +139,15 @@ def __call__(self, *inputs: torch.Tensor) -> List[torch.Tensor]:
 
         # CUDA zero-copy path: pre-allocate output tensors and register their device
         # pointers so the final kernel writes directly into PyTorch's buffer.
+        #
+        # Only the dtypes luminal *natively* writes can zero-copy. float64 is
+        # collapsed to f32 internally; registering an f64 device-ptr would
+        # have the kernel write 12 bytes into a 24-byte buffer, leaving half
+        # of every f64 element as garbage. We pre-allocate the f64 tensor so
+        # the collection path can fill it via `get_output` + cast, but skip
+        # the device-ptr handoff.
```
asglover: This comment is strange and worrisome. Why is float 64 collapsed to F32 internally?

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -150,11 +181,32 @@ def __call__(self, *inputs: torch.Tensor) -> List[torch.Tensor]:
                     else torch.float32
                 )
                 out = output_tensors[i]
-                if out_dtype.is_floating_point:
+                if out_dtype in _zero_copy_native_floats:
                     if not self._graph.output_is_zero_copy(name):
                         self._graph.copy_output_to_device_ptr(
                             name, out.data_ptr(), out.numel() * out.element_size()
                         )
+                elif out_dtype == torch.float64:
```
asglover: Why do these have to be special cases when the other dtypes are not?

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -188,7 +240,16 @@ def __call__(self, *inputs: torch.Tensor) -> List[torch.Tensor]:
                     if i < len(output_dtype_codes)
                     else torch.float32
                 )
-                if out_dtype in _int_dtypes:
+                if out_dtype == torch.float64:
```
asglover: Why this special casing?

src/hlir.rs
```diff
@@ -1202,20 +1202,48 @@ impl NativeOp for Cast {
...
                 NativeData::Bf16(f) => f.iter().map(|f| f.to_f32()).collect(),
                 NativeData::Int(i) => i.iter().map(|i| *i as f32).collect(),
+                NativeData::I64(i) => i.iter().map(|i| *i as f32).collect(),
+                NativeData::F64(f) => f.iter().map(|f| *f as f32).collect(),
+                NativeData::Bool(b) => b.iter().map(|b| if *b { 1.0 } else { 0.0 }).collect(),
+            }),
+            DType::F64 => NativeData::F64(match &input[0] {
+                NativeData::F64(f) => f.clone(),
+                NativeData::F32(f) => f.iter().map(|f| *f as f64).collect(),
+                NativeData::F16(f) => f.iter().map(|f| f.to_f32() as f64).collect(),
```
asglover, line 1212: Why are we implicitly converting?

src/hlir.rs
```diff
@@ -2728,6 +2835,8 @@ impl NativeData {
             NativeData::F32(v) => f16::from_f32(v[i]),
             NativeData::Bf16(v) => f16::from_f32(v[i].to_f32()),
             NativeData::Int(v) => f16::from_f32(v[i] as f32),
+            NativeData::I64(v) => f16::from_f32(v[i] as f32),
```
asglover: cast?

src/hlir.rs
```diff
@@ -2760,6 +2873,47 @@ impl NativeData {
         }
     }
 
+    /// 64-bit signed integer accessor. Used by I64-aware kernels; widens other
```
asglover: why does this exist? shouldn't casts be explicit? What is an I64 aware kernel?

src/hlir.rs
```diff
@@ -2768,6 +2922,8 @@ impl NativeData {
             NativeData::F16(v) => v[i].to_f32() != 0.0,
             NativeData::Bf16(v) => v[i].to_f32() != 0.0,
             NativeData::Int(v) => v[i] != 0,
+            NativeData::I64(v) => v[i] != 0,
+            NativeData::F64(v) => v[i] != 0.0,
```
asglover: What is NativeData and why is it in HLIR? (this might be too general a question for this PR.) Maybe something to address a different time.

crates/luminal_cuda_lite/src/runtime.rs
jafioti: agreed we should error if the value is not i64

crates/luminal_python/rust/src/typed_data.rs
jafioti: i think we should just support the i16, u8, i8 natively. if they arent supported, just error for now.

tucker-luminal:
The failing tests is to do to the modal runner for that test asking for a "--a100" which modal can choose to return as the 80GB modal(where the test will pass) or the 40 GB(where it will fail)

jafioti:
there should be a way to choose an 80gb a100: https://modal.com/docs/guide/gpu there is an option for `A100-80GB`

tucker-luminal:
> there should be a way to choose an 80gb a100: https://modal.com/docs/guide/gpu there is an option for `A100-80GB`

Yeah there is an update in this PR to change that over going forward, just the normal problem of it needs to be merged into main to fix :|

crates/luminal_python/rust/src/pt2_util.rs
```diff
@@ -198,16 +198,33 @@ pub fn resolve_neg1_dim_exprs(
     }
 }
 
-/// Map torch dtype integer (PT2 format) to luminal DType.
-/// PT2 numbering: 1=uint8, 2=int8, 3=int16, 4=int32, 5=int64, 6=float16, 7=float32, 8=float64, 12=bool, 13=bfloat16
+/// Map a PT2 dtype code to luminal `DType`.
+///
+/// Narrow ints (`Byte` / `Char` / `Short`) currently collapse to
```
asglover: Do we really not have byte, char, short? Is short not an int16? I guess char and byte should both be the same? And both be uint8? Can we just add these to our type list, even if we don't support them, just because we probably should have them anyways.

asglover: I'm okay with not having complex numbers for now or some of the others.

crates/luminal_python/rust/src/pt2_util.rs
```diff
@@ -198,16 +198,33 @@ pub fn resolve_neg1_dim_exprs(
...
-        8 => DType::F32, // float64 → F32 (no F64 in luminal)
-        13 => DType::Bf16,
-        12 => DType::Bool,
-        1..=5 => DType::Int, // uint8, int8, int16, int32, int64
-        _ => DType::F32,
+    let t = crate::torch_dtype::TorchDType::from_code(dtype)
+        .unwrap_or_else(|c| panic!("torch_dtype_int_to_luminal: unknown PT2 dtype code {c}"));
+    match t {
+        // Narrow ints collapse to Int (see doc above).
+        crate::torch_dtype::TorchDType::Byte
```
asglover, line 210: Same, I'd rather just have types for these or reject vs doing this dtype mapping.

crates/luminal_python/rust/src/torch_dtype.rs
```diff
@@ -0,0 +1,215 @@
...
+//! The PT2 export pipeline wire-serializes tensor dtypes as `u32` codes drawn
+//! from `torch._export.serde.schema.ScalarType` (an `IntEnum` on the Python
+//! side). Three sites in this crate used to carry duplicate raw-`u32` match
+//! arms with the canonical numbering hand-rolled in each — silent miscompile
+//! risk when PyTorch renumbers or adds a code. This module collapses those
+//! sites onto one typed enum and pins the numbering with a parity test that
+//! asserts every Rust variant matches `torch._export.serde.schema.ScalarType`
+//! at CI time (see `crates/luminal_python/tests/test_torch_dtype_parity.py`).
+//!
+//! Note: PyTorch's C++ `c10::ScalarType` uses a different numbering than the
```
asglover, line 12: Is this the only difference? I really hope that they don't differ in significant ways, that would be insane. If you could accidentally pass one enum, have it be interpreted as the other, and there would be a mismatch.

crates/luminal_python/rust/src/torch_dtype.rs
```diff
@@ -0,0 +1,215 @@
...
+    BFloat16 = 13,
+    Uint16 = 28,
+    Float8E4m3Fn = 29,
+    Float8E5m2 = 30,
+    Float8E4m3Fnuz = 31,
+    Float8E5m2Fnuz = 32,
+}
+
+impl TorchDType {
+    /// All variants, in declaration order. Used by the pyo3-exported parity
```
asglover, line 46: This is great! Thanks for tackling this.

crates/luminal_python/rust/src/torch_dtype.rs
```diff
@@ -0,0 +1,215 @@
...
+            DType::F32 | DType::TF32 => TorchDType::Float,
+            DType::F64 => TorchDType::Double,
+            DType::F16 => TorchDType::Half,
+            DType::Bf16 => TorchDType::BFloat16,
+            DType::Int => TorchDType::Int,
+            DType::I64 => TorchDType::Long,
+            DType::I8 => TorchDType::Char,
+            DType::U8 => TorchDType::Byte,
+            DType::I16 => TorchDType::Short,
+            DType::U16 => TorchDType::Int, // pre-existing workaround; see doc above.
```
asglover: This should not be, as above comments.

crates/luminal_python/rust/src/torch_dtype.rs
tucker-luminal: since we don't consume the pytorch directly, we go through the pt export it makes sense to be based off of the schema. If we ever moved off of that we would have to rethink this, but that would be a larger scale rewrite already

crates/luminal_python/rust/src/torch_dtype.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+}
+
+/// luminal `DType` → PyTorch dtype. `Err(dtype)` for luminal-specific
+/// variants without a PyTorch counterpart (`I4`, `U4`, `F6E2M3`, ...).
+///
+/// `DType::TF32` maps to `Float` (PyTorch's TF32 is a compute mode of f32,
+/// not a separate dtype; storage is f32).
+///
+/// `DType::U16` historically mapped to PyTorch's `Int` (PT2 code 4) as a
```
asglover: maybe the same comment

crates/luminal_python/rust/src/torch_dtype.rs
```diff
@@ -0,0 +1,215 @@
...
+            | TorchDType::Float8E5m2Fnuz => return Err(t),
+        })
+    }
+}
+
+/// luminal `DType` → PyTorch dtype. `Err(dtype)` for luminal-specific
+/// variants without a PyTorch counterpart (`I4`, `U4`, `F6E2M3`, ...).
+///
+/// `DType::TF32` maps to `Float` (PyTorch's TF32 is a compute mode of f32,
+/// not a separate dtype; storage is f32).
```
asglover: We probably shouldn't match this at all, because we never really should match this direction. it should just error. If we ever do have a TF32 in the wild and try to push it to pytorch that would be an unusual situation. we might want to just error. because pytorch can't actually handle TF32. We would need actual cast to FP32.

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -114,8 +137,57 @@ def __call__(self, *inputs: torch.Tensor) -> List[torch.Tensor]:
 
         output_dtype_codes = self._graph.output_dtypes
 
-        # CUDA zero-copy path: pre-allocate output tensors and register their device
-        # pointers so the final kernel writes directly into PyTorch's buffer.
+        # Per-dtype dispatch table: `{torch_dtype: (getter_name,
+        # read_dtype, final_cast_or_None)}`. `getter_name` is the
+        # `_graph` method that returns the buffer at `read_dtype`;
+        # `final_cast` is non-None only when the kernel emits an i32
```
asglover: this seems a little strange, are we making vectors that are in32 and putting narrow int types in each slot? What is going on?

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -114,8 +137,57 @@ def __call__(self, *inputs: torch.Tensor) -> List[torch.Tensor]:
...
+            torch.int16: ("get_output_i32", torch.int32, torch.int16),
+            torch.uint8: ("get_output_i32", torch.int32, torch.uint8),
+            torch.bool: ("get_output_bool", torch.bool, None),
+        }
+
+        def _read_typed_output(name: str, shape, out_dtype) -> torch.Tensor:
+            """Pull one output back from the runtime at the right dtype.
+            Routes through the per-dtype dispatch table; for any float
+            dtype not in the table, falls back to the generic f32
+            getter and post-casts."""
```
asglover: Why do we fallback to f32 instead of erroring?

crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -126,86 +198,29 @@ def __call__(self, *inputs: torch.Tensor) -> List[torch.Tensor]:
...
-                    code_to_torch_dtype(output_dtype_codes[i])
-                    if i < len(output_dtype_codes)
-                    else torch.float32
-                )
+        outputs = []
+        for i, (name, shape) in enumerate(zip(self._output_names, output_shapes)):
+            out_dtype = (
+                code_to_torch_dtype(output_dtype_codes[i])
+                if i < len(output_dtype_codes)
+                else torch.float32
```
asglover: probably should error vs assume?

crates/luminal_python/src/luminal/dtype_util.py
```diff
@@ -1,26 +1,44 @@
...
+    torch.bfloat16: ScalarType.BFLOAT16.value,
 }
 
 _CODE_TO_TORCH_DTYPE = {v: k for k, v in _TORCH_DTYPE_TO_CODE.items()}
 
 
 def torch_dtype_code(dtype):
     """Map torch.dtype to PT2 dtype integer code."""
-    return _TORCH_DTYPE_TO_CODE.get(dtype, 7)  # default to f32
+    return _TORCH_DTYPE_TO_CODE.get(dtype, ScalarType.FLOAT.value)
```
asglover: should we have a default here? or should we error?

crates/luminal_python/tests/test_dtype_boundary.py
```diff
@@ -0,0 +1,249 @@
...
+    device_name = request.param
+    if device_name == "cuda":
+        skip_reason = _cuda_skip_reason()
+        if skip_reason is not None:
+            pytest.skip(skip_reason)
+    return torch.device(device_name)
+
+
+# Dtypes that round-trip the BoundaryNoopModel without an explicit
+# `x.to(model.input_dtypes[0])` cast at the call site. Anything not in this
```
asglover, line 124: this bit needs to be updated when we just error instead of widening/ narrowing.

crates/luminal_python/tests/test_dtype_boundary.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+@pytest.mark.parametrize(
+    "case",
+    [
+        pytest.param(case, id=case.name)
+        for case in DTYPE_CASES
+        # Narrower integer widths still collapse to luminal's `Int` (i32) at
+        # the boundary; user inputs of those dtypes don't match the graph's
+        # declared input dtype and so trigger a hard reject. int64 / float64
```
asglover: again needs updating

crates/luminal_python/tests/test_dtype_boundary.py
```diff
@@ -0,0 +1,249 @@
...
+        if case.name
+        in {
+            "bool",
+            "int32",
+            "float16",
+            "bfloat16",
+            "float32",
+            # int64 / float64 are first-class in the IR — passing a tensor
+            # of either dtype matches the graph's input dtype directly, no
+            # conversion needed.
```
asglover, line 219: some comments like this can be removed, there are a few other locations.

src/hlir.rs
asglover: It is still strange to cast to an f32 then interpret as an F64?

src/hlir.rs
```diff
@@ -1202,20 +1202,49 @@ impl NativeOp for Cast {
...
             DType::Int => NativeData::Int(match &input[0] {
                 NativeData::F32(f) => f.iter().map(|f| *f as i32).collect(),
                 NativeData::F16(f) => f.iter().map(|f| f.to_f32() as i32).collect(),
                 NativeData::Bf16(f) => f.iter().map(|f| f.to_f32() as i32).collect(),
                 NativeData::Int(i) => i.clone(),
+                // Saturating `Cast(I64 -> Int)`. This is the explicit
+                // graph node — `Cast` IS the user/translator-emitted
+                // operation, not an implicit bridge. Values outside the
+                // i32 range wrap via Rust's `as i32`, matching
+                // `tensor.to(torch.int32)` semantics on overflow.
```
asglover, line 1227: okay, this makes sense, but the comment can be removed.

src/hlir.rs
```diff
@@ -1202,20 +1202,49 @@ impl NativeOp for Cast {
...
+                NativeData::F16(f) => f.iter().map(|f| f.to_f32() as i64).collect(),
+                NativeData::Bf16(f) => f.iter().map(|f| f.to_f32() as i64).collect(),
                 NativeData::Bool(b) => b.iter().map(|b| if *b { 1 } else { 0 }).collect(),
             }),
             DType::F16 => NativeData::F16(match &input[0] {
                 NativeData::F32(f) => f.iter().copied().map(f16::from_f32).collect(),
                 NativeData::F16(f) => f.clone(),
                 NativeData::Bf16(f) => f.iter().map(|f| f16::from_f32(f.to_f32())).collect(),
                 NativeData::Int(i) => i.iter().map(|i| f16::from_f32(*i as f32)).collect(),
+                NativeData::I64(i) => i.iter().map(|i| f16::from_f32(*i as f32)).collect(),
```
asglover, line 1246: why do these casts between I32 and F64 and all these other things go through F32 or I32 first? Shouldn't they be "directly" translated and rounded or truncated if needed in one shot? I'm worried that this could be a source of some subtle rounding bug eventually.

src/hlir.rs
```diff
@@ -1259,8 +1292,22 @@ fn unary_impl(
         NativeData::F32(f) => NativeData::F32(ind.map(|i| f32_fn(f[i])).collect()),
         NativeData::F16(f) => NativeData::F16(ind.map(|i| f16_fn(f[i])).collect()),
         NativeData::Bf16(f) => NativeData::Bf16(ind.map(|i| bf16_fn(f[i])).collect()),
-        NativeData::Int(_) => panic!("not implemented for int"),
-        NativeData::Bool(_) => panic!("not implemented for bool"),
+        NativeData::Int(_) => panic!("unary_impl: no Int kernel — cast to F32 at the call site"),
```
asglover, line 1295: maybe these errors could be unified or simplified and comments removed

src/hlir.rs
```diff
@@ -2854,6 +2999,16 @@ impl From<Vec<i32>> for NativeData {
         NativeData::Int(value)
     }
 }
+impl From<Vec<i64>> for NativeData {
+    fn from(value: Vec<i64>) -> Self {
+        NativeData::I64(value)
+    }
+}
+// No `From<Vec<f64>> for NativeData` impl. Adding it makes plain
```
asglover, line 2951: Why don't we force every test site to spell out: 

 forces every test
// site to spell out `Vec::<f32>::from([...])`.

This seems more sound?

src/hlir.rs
asglover, line 2874: Why do we provide this casting at in the interface at all? Instead of only giving user the actual datatype? I get that it might be convenient at some test site to upcast or something, so that narrow dtypes could be compared in FP32 or something, but why doesn't that upcasting code live in the test suite instead of in the native front end. It sort of short cuts a lot of our type safety.

PR #326, opened 2026-05-19
Optimize Metal runtime setup and buffer reuse

spinlocked:
Cached decode repeatedly executes the same bucketed Metal graph with changing dynamic values. These changes reduce repeated execution overhead by avoiding unnecessary intermediate-buffer reallocations, MPS descriptor/kernel setup, and per-run execution metadata reconstruction.

In the Qwen Metal example on my M3 Max, this improves TPOT by ~2%.

Reuse dynamic intermediate buffers at bucket capacity
  - Metal intermediate buffers are allocated at the active bucket's max dynamic dimensions.
  - The runtime tracks each buffer's logical byte length separately, so smaller executions inside the same bucket can safely reuse the larger allocation.
  - `get_f32`, buffer clearing, bucket activation, and `remove_buffer` now keep that logical-length metadata consistent.

Cache MPS matmul setup objects
  - Added an `MpsKernelCache` for `MPSMatrixDescriptor` and `MPSMatrixMultiplication` objects.
  - Introduced `MetalEncodeContext` so kernels receive the command buffer, dynamic-dimension buffer, and MPS cache together.
  - `MPSMatmul` and `MPSBatchedMatmul` now reuse cached descriptors and multiplication kernels instead of recreating them on every encode.

Precompute per-bucket execution metadata
  - Bucket compilation now records LLIR input mappings, output data mappings, and executable kernel steps.
  - `execute` and `execute_timed` reuse this precomputed execution plan instead of rebuilding topological order, input node lists, and dtype metadata every run.

### DIFF
--- a/crates/luminal_metal/src/kernel/mod.rs
+++ b/crates/luminal_metal/src/kernel/mod.rs
@@ -6,9 +6,126 @@ pub use ops::*;
 use luminal::dtype::DType;
 use luminal::op::EgglogOp;
 use luminal::prelude::*;
-use metal::{Buffer, CommandBufferRef, ComputeCommandEncoderRef, ComputePipelineState, Device};
+use metal::{
+    Buffer, CommandBufferRef, ComputeCommandEncoderRef, ComputePipelineState, Device,
+    foreign_types::ForeignTypeRef, mps,
+};
+use objc::rc::StrongPtr;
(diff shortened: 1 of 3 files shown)


jafioti:
awesome I'm getting #327  merged right now and then I'll get this and #328  merged

PR #331, opened 2026-05-21
Tucker/llama3 rsqrt fix

tucker-luminal:
Ensure that we use the rsqrtf fusion
Change how we lower pox(x, 2)


tucker-luminal:
The failing tests is do to that fact that Qwen3 Moe 30B can't run on a A100. The modal run uses --a100 to request which card it wants. Modal can promote that from the 40GB version(where it will fail) to the 80GB version where it will work. 

https://modal.com/docs/guide/gpu

PR #332, opened 2026-05-22
Add an opt-in graph search compile cache

spinlocked:
Add a disk-backed compile cache that persists the selected e-graph choices produced by graph search. Cache entries store serialized e-graphs plus the winning reachable choice set for each dynamic-dimension bucket, then rebuild and load LLIR from those choices on a cache hit. This supports per-bucket saturation and skips repeated e-graph saturation/search/profiling work.

Integrate automatic cache lookup into the generic Graph::compile and Graph::search paths. Automatic caching remains off by default and is enabled with LUMINAL_COMPILE_CACHE=1. Cache keys include the runtime type, graph source fingerprint, dynamic dims, dim intervals, dim buckets, compile options, and search-space memory cap so stale selections are ignored when relevant compile inputs change.

Make SerializedEGraph serde-serializable for cache storage and add regression tests that write cache entries, reload them on a second search, and verify the cached path does not profile candidates again.

### DIFF
--- a/src/egglog_utils/mod.rs
+++ b/src/egglog_utils/mod.rs
@@ -348,7 +348,7 @@ use crate::{
 use egglog::{ArcSort, CommandOutput, EGraph, Value};
 use egglog_reports::ReportLevel;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 ///  This is snapshot of an EGraph with Rust native hash maps and sets for enabling more native traversal / algorithm writing.
 ///  The name comes from the serialize egraph crates, which returns a ETermDAG, which caused issues, so this is a homebrew semi-static egraph
 pub struct SerializedEGraph {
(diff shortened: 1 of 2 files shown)


jafioti:
One big change just merged in with main, we rerun e-graph saturation for each bucket, rather than just use the same e-graph and do seperate searches per bucket. So we'll need to save multiple e-graphs now.

The reason for this has to do with dyn dim intervals. We can now do analyses that involve the intervals a dyn dim will inhabit, which lets us make guarentees like "we know that sequence length will always be below 2048" in the actual egglog rules.

spinlocked:
Ok, interesting. I'll look into updating this to be compatible with those changes when I can.

spinlocked:
@jafioti I've updated this to be compatible with the change you landed on main. The main adjustment was updating the compile cache to persist the same per-bucket e-graph structure that search now builds: one serialized e-graph per bucket combination, with each cached choice set tied to its matching e-graph.

I see the caching work correctly on my local MBP when running the following twice: `LUMINAL_COMPILE_CACHE_DIR=/tmp/luminal-qwen-cache LUMINAL_COMPILE_CACHE=1 cargo run --release -p qwen --features metal`

PR #341, opened 2026-06-01
cuda_lite: bf16 NVRTC ambiguity fixes (Recip and reduce_max_k)

tucker-luminal:
## Summary

- `region_codegen.rs::Recip`: cast the `1.0f` literal to the input dtype so `/` resolves unambiguously for bf16 inputs (NVRTC otherwise can't pick between `operator/(bf16, bf16)` and `operator/(float, float)`).
- `hlir.rs::reduce_max_k`: cast `NEG_INF_F` to the input dtype at both use sites (initializer + warp-reduce ternary) for the same reason.

Two files, three line changes total.

## Test plan

- [x] `./run_tests_cuda.sh --include-slow` — 494 passed, 2 skipped, 4 xfailed, 23 min
- [x] Build clean on CUDA 12.8

### DIFF
--- a/crates/luminal_cuda_lite/src/kernel/fusion/region_codegen.rs
+++ b/crates/luminal_cuda_lite/src/kernel/fusion/region_codegen.rs
@@ -362,7 +362,7 @@ fn elementwise_body(op: &str, locals: &[&str], dtype: DType) -> String {
         "Exp" => format!("expf({})", a()),
         "Exp2" => format!("exp2f({})", a()),
         "Log2" => format!("log2f({})", a()),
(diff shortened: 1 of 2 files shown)


tucker-luminal:
This showed up when trying to get Gemma working through luminal python, just small issues with how we handle bf16 in cuda

tucker-luminal:
The failing cuda test is a flaky one, it's a llama decode test where some times the error is 5e-5 ish vs the tolerance of 1e-5. It is unrelated to this set of changes, but I will look into it if we think that that is too much

PR #342, opened 2026-06-01
Tucker/cublaslt fp8 guard fast fail

tucker-luminal:
One of the fixes for Gemma4_moe through luminal_python, we OOM due to f8 matching rules(which there are not f8 ops in the graph). This is temp fix, that @asglover and I talked about in person as well.

### DIFF
--- a/crates/luminal_cuda_lite/src/host/cublaslt/cublaslt_fp8_rewrite.egg
+++ b/crates/luminal_cuda_lite/src/host/cublaslt/cublaslt_fp8_rewrite.egg
@@ -10,6 +10,7 @@
 
 (rule
     (
+        (cublaslt_fp8_present)
         ; Match the scaled FP8 linear form directly before the unscaled FP8
         ; matmul rewrite can hide the quantize/dequant scale structure.
         (= ?scaled_activation (Op (Mul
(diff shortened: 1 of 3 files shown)


tucker-luminal:
@asglover and @jafioti This is a temporary fix to help luminal_python process gemma4_moe. 
Basic idea is that the f8 cublastlt rules are matching so much that they blow out the memory on a gh box(which has something like 450 GB of combined memory). Talked to Austin in person about this problem and the fix as well and he felt it was an ok fix to get in for now, since the real fix would require rewrite these rules, which is a project unto itself, and he said that likely won't happen until this current work on input/output modeling(the manifesto) is finished

tucker-luminal:
The failing cuda test is a flaky one, it's a llama decode test where some times the error is 5e-5 ish vs the tolerance of 1e-5. It is unrelated to this set of changes, but I will look into it if we think that that is too much

jafioti:
Tbh I really dont like this, I think the rule should just be actually altered. I think the alteration would be fairly minor. Yes things will change a lot when the new hlir comes in, but having everything working on the old hlir will make the transition much easier, rather than have to also fix other broken things that were bandaided. If a proper fix cant be done, we should leave it failing for now.

jafioti:
The goal is not to get tests passing, the goal is to have a correct codebase, which should incidentally make the tests pass.

asglover:
That is a reasonable position, but tucker wasn't in a position to evaluate why this rule is blowing up, not having the context. I could not by inspection find what was wrong. we'll need to do a closer look at the egglog rules to try and understand what is causing the blow up.

PR #344, opened 2026-06-04
CublasLT and flashinfer in cuda graphs, plus mem limits

jafioti:
This PR puts cublasLT and flashinfer ops into cuda graphs and surgically modifies them when dyn dims change. Significantly improves TPOT on dense models not much difference in MoE models (I suspect because the fused MoE ops are still host ops).

It also adds a much better memory estimator (rust-side) to prevent OOMs in a fairly generic pattern where backends can add their own LLIR filters.


crates/luminal_cuda_lite/src/host/cublaslt/mod.rs
```diff
@@ -1126,6 +1383,171 @@ impl CuBlasLt {
...
+            "cuBLASLT_resolve_graph",
+            m, n, k, lda, ldb, ldc, ldd, batch_count, ?a_layout, ?b_layout,
+            ?self.a_order, ?self.b_order, ?self.c_order, ?self.d_order,
+            ?self.a_dtype, ?self.b_dtype, ?self.c_dtype, ?self.d_dtype,
+            ?self.compute_type, ?self.scale_dtype, self.alpha, self.beta,
+            ?self.epilogue,
+        )
+        .entered();
+
+        const WORKSPACE_SIZE: usize = 32 * 1024 * 1024;
```
asglover, line 1484: Where does this come from?

crates/luminal_cuda_lite/src/host/cublaslt/mod.rs
```diff
@@ -796,12 +961,15 @@ fn set_scalar_scale_pointer(
     Ok(())
 }
 
-fn run_cublaslt_matmul(
+pub(crate) fn prepare_cublaslt_matmul(
```
asglover, line 964: Not that it exists currently, but this kind of prepare and then pass an object at runtime pattern would be a good thing to have in the compiler. We'll need it in other places. But this is a fine solution for now.

crates/luminal_cuda_lite/src/host/cublaslt/mod.rs
jafioti: I believe that is the max workspace cublasLT needs. Found through a trial and error process. I agree it would be nicer to be more principled.

out.png
asglover, line 1: did you mean to commit this?

jafioti: Yeah should be removed now

PR #351, opened 2026-06-13
Moe bf16 flashinfer pure hlir

jafioti:


tucker-luminal:
Can you check this PR against the Fp8 blow up rules I sent you, you are working out of one of the file that has the rules that caused the problem and I am curious if you happened to solved/lessened the problem incidentally

PR #357, opened 2026-06-22
perf+fix(loop-rolling): make the auto-roll prepass tractable + cycle-safe on large graphs

jafioti:
The auto-loop-rolling prepass (folds repeated structure, e.g. transformer layers, into a loop) had perf and correctness problems on large real graphs (a 36-layer MoE forward is ~3.4k HLIR nodes/layer; the dual-branch minibatch-prefill variant ~6.7k). Fixes:

Perf (the prepass took 16+ minutes; now seconds):
- Memoize per-node rolling signatures. `rolling_op_signature` Debug-formats each op (allocating + recursing over shape/stride metadata) and the search calls it for the same node across many overlapping windows; that re-formatting dominated. Split into a cached wrapper + `_uncached` impl, keyed by NodeIndex (pure while the graph is read-only during search), cleared per `best_rolling_candidate`.
- Fix `canonicalize_occurrence`'s output-node scan: `graph.externals(Outgoing)` re-scanned EVERY node for each node in the window (O(window × graph)); check the node's own out-edges instead (O(out-degree)).
- Cap the largest probed window (LUMINAL_MAX_ROLL_BODY, default 8192). With two structurally-similar branches the cheap rolling hash matches many huge windows that each trigger an O(window) canonicalize that ultimately fails; capping above one layer skips them without changing the selected candidate.

Correctness (prevents a panic):
- `candidate_is_rollable`: reject candidates whose non-state boundary input is produced by ANOTHER occurrence (a non-adjacent cross-occurrence dep, e.g. occ `i` consuming occ `i-2`). Folding such bodies into one creates a directed cycle, which makes the egglog Kahn toposort drop nodes and panic. Applied at seed selection and after `grow_rolling_candidate` (growing can re-introduce the dep). Unit-tested.

Plus escape hatch LUMINAL_DISABLE_AUTO_LOOP_ROLLING and LUMINAL_DEBUG_ROLL_PROGRESS diagnostics. Behavior-preserving for the common case (126 lib tests pass).

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -896,6 +896,9 @@ impl Graph {
     }
 
     fn best_rolling_candidate(&self, max_region_size: usize) -> RollingSearchReport {
+        // The signature memo is keyed by NodeIndex; clear it so entries from a
+        // prior (now-mutated) graph state can't leak into this read-only search.
+        clear_rolling_sig_cache();
         let Some(full_topo) = stable_toposort_by_node_index(&self.graph) else {
             return RollingSearchReport {
                 candidate: None,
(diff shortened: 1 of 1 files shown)


jafioti:
### The failing **Python CUDA Tests** check is pre-existing on `main`, not introduced by this PR

I dug into the CI logs. This PR's failing `Python CUDA Tests` run reports **4 failed / 488 passed**:

- `test_hlir_ops.py::test_gather_negative_indices`
- `test_hlir_ops.py::test_gather_2d_axis0`
- `test_qwen3_moe.py::test_hf_qwen3_moe_tiny` (`max_diff=4.59e-2`)
- `test_qwen3_moe.py::test_hf_qwen3_moe_medium` (`max_diff=3.32e-1`)

Every one of these has been failing on **`main` itself** since `205bd838` ("Moe bf16 flashinfer pure hlir", #351) — `Test Python CUDA` went red there and has stayed red (`64fa8d99` was the last green run). The *which-subset-fails* shifts run-to-run, which is the signature of CUDA bf16 numeric-tolerance flakiness rather than a deterministic regression:

| failing test | this PR | main `767e34be` (PR base) | `8adb576a` | `dc7e06d2` | `205bd838` |
|---|:--:|:--:|:--:|:--:|:--:|
| `test_gather_negative_indices` | ✗ | – | ✗ | ✗ | ✗ |
| `test_gather_2d_axis0` | ✗ | – | ✗ | – | – |
| `qwen3_moe_tiny` | ✗ | – | – | ✗ | ✗ |
| `qwen3_moe_small` | – | ✗ | ✗ | ✗ | – |
| `qwen3_moe_medium` | ✗ | ✗ | – | ✗ | ✗ |
| **total failed** | **4** | 2 | 3 | 4 | 3 |

This PR's failure set is a subset of the same pool; it adds **no new failing tests**.

Why this isn't the loop-rolling change:

- **CPU passes.** The gather tests live in `test_hlir_ops.py`, which the **Python Native (CPU)** job runs and passes. Loop rolling is a backend-agnostic graph transform — if it broke gather, CPU would fail too.
- **The real graphs are green.** Every Modal model example (llama / qwen / **qwen3_moe** / whisper / gemma / gemma4_moe on A100) — which exercise loop rolling on full layer stacks — passes.
- The failures are `torch.allclose` tolerance misses on MoE logits / CUDA gather edges, consistent with the `#351` bf16+flashinfer numeric drift.

**Net:** the red `Python CUDA Tests` check is inherited from a base that was already failing these flaky tests; Core/Cuda/Metal unit tests, Python **Native**, and all model examples are green here. The flaky `qwen3_moe`/`gather` CUDA tests look like a separate upstream `#351` issue (tolerance bump or `xfail`) and shouldn't block this PR.

PR #361, opened 2026-07-08
hlir: serialize egglog constants round-trip exact ({:.6} -> {:?})

tucker-luminal:
Change how we serialize constants into egglog

previously we where using .6, and so was losing some decimal points. 

Changes to just serialization the entire constant

updated some egglog rules where we look for exact constants to make sure they still match

In support of imagegen pipelines, broken out into its own PR because it is an issue unto itself


tucker-luminal:
Waiting for https://github.com/luminal-ai/luminal/pull/363 to get in and it contains a fix for a flaky test failing

PR #369, opened 2026-07-17
Add ATen bitwise ops for GPTQ graph lowering

jafioti:


jafioti:
unclear if this is actually needed / the right way to go right now

PR #378, opened 2026-07-22
egglo gemma 4 blow up fix

tucker-luminal:


tucker-luminal:
I have it as a PR so I can check the Ci/CD timings with this change in, not meant to be a PR yet

PR #379, opened 2026-07-23
Support stateful PT2 execution for demos

matthewjgunton:
## What changed

- Added PT2 translation support for StaticCache-style optional-tensor `index_put` operations and negative slice bounds.
- Preserved input-backed write-backs so mutable backend state can remain authoritative.
- Added a PyTorch/PT2 regression test in `crates/luminal_python/tests/test_hlir_ops.py` that exports a real `torch.nn.Module` through `torch.export`.
- Corrected loop-output resolution when an output aliases loop-carried state.
- Accepted symbolically equivalent expansion dimensions after simplification.
- Allowed backends to leave device-resident write-back outputs unmaterialized.
- Added opt-in input skipping for backend-owned persistent state.

## Why

Stateful compiled generation needs the KV cache to remain resident and mutable in the backend across prefill and decode calls. The previous PT2 boundary treated cache mutation as ordinary functional input/output traffic and did not translate every shape, indexing, and loop-aliasing form emitted by the demo graph.

The regression test now exercises the same PyTorch-to-PT2 export boundary used by real applications instead of manually constructing a serialized ATen representation.

## Impact

This enables stateful compiled demos to preserve backend-owned cache state while retaining PyTorch's functionalized output contract. Input skipping remains opt-in, so existing backends continue sending every input by default.

## Validation

- No branch-added comments or debug environment-variable hooks remain.
- `cargo fmt --all -- --check`
- `cargo test -p luminal_python` — 13 passed
- `uv run pytest tests/test_hlir_ops.py::test_input_backed_index_put_output_aliases_destination_input -q` — 1 passed
- Ruff 0.14.5 format and lint checks for `tests/test_hlir_ops.py`
- `ruff format --check crates/luminal_python/src/luminal/compiled_model.py`
- `ruff check crates/luminal_python/src/luminal/compiled_model.py`
- `cargo clippy --workspace --exclude luminal_cuda_lite --exclude luminal_metal --exclude luminal_bench --all-targets -- -D warnings`
- `cargo test -p luminal --lib` — 121 passed
- Previously validated end-to-end with the Luminal Qwen3-MoE chat demo using a retained compiled artifact across generation.


crates/luminal_python/src/luminal/compiled_model.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         self._has_dynamic_dims = getattr(graph_result, "has_dynamic_dims", False)
         self._weight_refs = weight_refs or []
         self._user_indices = user_indices
+        self.skip_input_names = frozenset()
```
tucker-luminal, line 48: Does this get written to from some where?

crates/luminal_python/rust/src/translator/mod.rs
```diff
@@ -400,3 +390,105 @@ impl<'a> Translator<'a> {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use serde_json::json;
+
+    use super::*;
+
+    fn input_backed_index_put_program() -> ParsedPT2 {
```
tucker-luminal: Please express this as a pytorch model that gets compiled through luminal, it helps make sure we are not to keyed off a particular version of PT

PR #384, opened 2026-07-27
translate_sdpa: close the SDPA surface — precision, masks, GQA, tuple…

tucker-luminal:
Improving our SDPA (scaled dot product attention) aten op handling. This is a well used op for pytorch when it comes to attention so went with broad coverage on this.


tucker-luminal:
Putting this up so I can do some timings on it before I put it up for full review

PR #387, opened 2026-08-03
Improving our d-type promotion to match pytorch accumulation behavior + fix bug in expand op

tucker-luminal:
Better following the bf16 accumulation behavior in pytorch, following the behavior laid out by https://docs.pytorch.org/docs/stable/tensor_attributes.html#type-promotion-doc), and https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/OpMathType.h

OpMath is name given to the in house style of pytorch for how they handle type promotion to avoid accumulation errors


jafioti:
looks good! feel free to merge once it's green

asglover:
nice work! It's a shame prim torch isn't real and there's not a way to get pytorch to tell you what it's doing. Great work digging in and figuring this promotion logic out!

PR #396, opened 2026-08-12
Changing symbolic dimensions from bare chars to a structure backed by a String

tucker-luminal:
We used to have a char back the symbolic dimensions. This left us with only those values we could fit into chat as names, it also meant we had a limited set of them, a->z, A->Z. With lowercase z actually reserved for the index dimension. On graphs like Qwen 3.5, where it has Gated Delta Attention, this function torch_recurrent_gated_delta_rule would produce more symbolic dimensions than we could support. 

https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_5/modeling_qwen3_5.py

So we are upgrading from being backed by a char to being backed by a string so we can have more symbolic names and those names can be more descriptive as well over time.

### DIFF
--- a/crates/luminal_cuda_lite/src/host/flashinfer/mod.rs
+++ b/crates/luminal_cuda_lite/src/host/flashinfer/mod.rs
@@ -44,6 +44,9 @@ pub struct FlashInferAttention {
     pub head_dim: usize,
     pub page_size: usize,
     pub batch_dim: Expression,
+    /// Total tokens attended over. Captured from the matched shape by the
+    /// rules, so it does not depend on what the graph calls this dim.
+    pub context_dim: Expression,
     pub dtype: DType,
     /// Softmax scale; 0.0 = default `1/sqrt(head_dim)`.
     pub sm_scale: f64,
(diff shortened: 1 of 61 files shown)


src/shape/expression.rs
```diff
@@ -361,12 +378,25 @@ impl Expression {
...
+    /// Substitutes on the term, so no dim name can be a prefix of another and
+    /// get rewritten mid-name.
+    pub fn to_kernel_with_index(&self, index_var: &str) -> String {
+        self.to_kernel_with(index_var, &kernel_const_name)
+    }
+
+    /// As [`to_kernel_with_index`](Self::to_kernel_with_index), but also chooses
+    /// how each dynamic dim is spelled — Metal names them by buffer slot rather
+    /// than by `const_<name>`.
+    pub fn to_kernel_with(&self, index_var: &str, dim: &dyn Fn(&Symbol) -> String) -> String {
```
tucker-luminal, line 393: the old one of doing this was the "to_kernal().replace()", this improves it, folding what was the replace into the function itself(the index_var.to_string()). This also makes an assumption, all of the places that where doing this where doing it over the "const_z", do this formalizes that as well, with the "c.is_reserved()" line. A fair point to argue would be that this specializes this function to much, but I think usage has showed that it really only had a single use.

crates/luminal_cuda_lite/src/host/flashinfer/mod.rs
```diff
@@ -272,6 +272,14 @@ unsafe impl Sync for PreparedFlashInferDecode {}
 unsafe impl Send for FlashInferAttention {}
 unsafe impl Sync for FlashInferAttention {}
 
+/// The context dim. The `.egg` rules match `(MVar "c")` literally, so this
+/// spelling is a contract with those files.
+pub(crate) const CONTEXT_DIM_NAME: &str = "c";
```
jafioti: What is this? This is special cased, we need to delete this and change the rewrite rule to be agnostic to dyn dim name

crates/luminal_cuda_lite/src/host/flashinfer/mod.rs
```diff
@@ -272,6 +272,14 @@ unsafe impl Sync for PreparedFlashInferDecode {}
...
+
+/// The indptr-rows dim, read from `dyn_map` at execute time. No `.egg` rule
+/// matches it.
+pub(crate) const INDPTR_ROWS_DIM_NAME: &str = "r";
```
jafioti: Same with this

PR #411, opened 2026-08-20
metal: add BF16 kernel and runtime support

spinlocked:
MPSMatmul currently supports only F32 and F16 storage, so restrict its rewrites to those dtypes and use the generic Metal matmul path for BF16.

### DIFF
--- a/crates/luminal_metal/src/kernel/ops.rs
+++ b/crates/luminal_metal/src/kernel/ops.rs
@@ -57,7 +57,7 @@ pub type MetalOps = (
 
 fn compile_shader(device: &Device, source: &str, function_name: &str) -> ComputePipelineState {
     let options = metal::CompileOptions::new();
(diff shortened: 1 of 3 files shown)


crates/luminal_metal/src/kernel/ops.rs
```diff
@@ -57,7 +57,7 @@ pub type MetalOps = (
 
 fn compile_shader(device: &Device, source: &str, function_name: &str) -> ComputePipelineState {
     let options = metal::CompileOptions::new();
-    options.set_language_version(MTLLanguageVersion::V2_4);
+    options.set_language_version(MTLLanguageVersion::V3_1);
```
spinlocked, line 60: Wasn't sure if this is OK. It's needed for `bfloat` support. Is there a reason I'm not aware of this to allow these older MSL versions?

PR #415, opened 2026-08-22
metal: guard every dtype source, not just Input nodes

BABTUNA:
## Problem

`reject_unsupported_dtype` only checks `Input` nodes. Dtypes also enter a graph through `Cast`, `ConstantF64`, and `CustomOpKind`, so those graphs slip past the guard and panic in codegen.

```rust
cx.constant_i64(1_234_567_890_123)
```

```
panicked at crates/luminal_metal/src/kernel/ops.rs:103:14:
Metal dtype Int64 is not supported yet
```

`constant_i64` builds its value from `Iota` limbs cast to I64, so the graph has no I64 `Input` at all. The dtype comes only from the casts, which the guard never looked at.

`cx.constant_float64(1.5)` fails the same way at `runtime.rs:692` with "cannot execute unlowered LLIR node".

## Fix

Check every op that introduces a dtype. Dtypes enter HLIR at four ops and propagate from there, so covering the sources covers the whole graph:

| op | dtype |
|---|---|
| `Input` | `.dtype` |
| `Cast` | target dtype |
| `ConstantF64` | F64 |
| `CustomOpKind` | `.dtype` |

The invariant is that no node may introduce a dtype Metal cannot emit, rather than that no input may. The propagated dtype facts live in the e-graph, not the Rust graph, so they cannot be walked here directly.

## Tests

| test | asserts |
|---|---|
| `metal_factory_rejects_i64_introduced_by_cast` | clean `Err` naming the dtype |
| `metal_factory_rejects_f64_constant` | clean `Err` naming the dtype |
| `metal_dtype_guard_accepts_supported_dtypes` | an F32/F16/Int graph is still accepted |

Both rejection tests panic on `main` and pass here. The third guards against the wider walk over-rejecting.

## Verified locally

M1 Pro, macOS 15:

```
cargo test --release -p luminal_metal -- --test-threads=1   58 passed, 0 failed
cargo clippy -p luminal_metal --all-targets                 clean
cargo fmt --all --check                                     clean
```

## Scope

This covers the `DynBackend` path, where the guard runs. `cx.compile::<MetalRuntime>(..)` called directly still panics, since adding a check there would change the core compile signature. Happy to follow up if you want that too.

### DIFF
--- a/crates/luminal_metal/src/dyn_backend.rs
+++ b/crates/luminal_metal/src/dyn_backend.rs
@@ -33,32 +33,60 @@ impl DynBackend for MetalDynBackend {
     }
 }
 
-/// Reject dtypes the Metal kernel emitters don't support.
+/// Dtypes the Metal kernel emitters cannot lower.
 ///
 /// Metal codegen has no native 64-bit integer or 64-bit float paths.
-/// Reaching the kernel emitter with one of these dtypes used to panic deep
-/// in MSL generation with an unhelpful error; surfacing a clean message
-/// at translate-time lets the user fall back to CPU or pick a narrower
-/// dtype before any Metal compilation runs.
-fn reject_unsupported_dtype(graph: &Graph) -> Result<(), String> {
+const UNSUPPORTED_DTYPES: [DType; 2] = [DType::I64, DType::F64];
+
+/// The dtype an HLIR node introduces into the graph, with a description of the
+/// node for diagnostics, or `None` if the node only propagates its inputs'
+/// dtypes.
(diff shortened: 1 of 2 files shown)


BABTUNA:
Closing as obsolete. The Metal crate rewrite removed `dyn_backend.rs` and `reject_unsupported_dtype`, and Metal now supports Int64 natively (`kernels.rs` maps `PlanDtype::Int64` to `long`, and the egglog rules reason about it directly), so the limitation this guarded against no longer exists.

